PR #113, opened 2015-05-14
Allow package listing to be sorted by Downloads or Name

parkerl:
@ericmj how does this look?

  - Issue #107

### DIFF
--- a/lib/hex_web/package.ex
+++ b/lib/hex_web/package.ex
@@ -135,9 +135,10 @@ defmodule HexWeb.Package do
     |> HexWeb.Repo.delete_all
   end
 
-  def all(page, count, search \\ nil) do
+  def all(page, count, search \\ nil, sort \\ :name) do
     from(p in HexWeb.Package,
-         order_by: p.name)
+         preload: :downloads)
+    |> sort(sort)
     |> Util.paginate(page, count)
     |> search(search, true)
     |> HexWeb.Repo.all
@@ -201,6 +202,16 @@ defmodule HexWeb.Package do
 
   defp like_escape(string, escape) do
     String.replace(string, escape, "\\\\\\1")
+  end
+
+  defp sort(query, :name) do
+    from d in query, order_by: :name
+  end
+
+  defp sort(query, :downloads) do
+    from p in query,
+    left_join: d in HexWeb.Stats.PackageDownload, on: p.id == d.package_id and d.view == "all",
+    order_by: [asc: is_nil(d.downloads), desc: d.downloads]
   end
 end
 
--- a/lib/hex_web/util.ex
+++ b/lib/hex_web/util.ex
@@ -20,6 +20,7 @@ defmodule HexWeb.Util do
     :calendar.gregorian_days_to_date(today_days - 1)
   end
 
+
   defp diff(a, b) do
     {days, time} = :calendar.time_difference(a, b)
     :calendar.time_to_seconds(time) - (days * 24 * 60 * 60)
--- a/lib/hex_web/web/router.ex
+++ b/lib/hex_web/web/router.ex
@@ -100,14 +100,15 @@ defmodule HexWeb.Web.Router do
     conn              = fetch_params(conn)
     packages_per_page = @packages
     search            = conn.params["search"] |> safe_query
+    sort              = safe_sort(conn.params["sort"] || "name")
 
     package_count = Package.count(search)
     page          = safe_page(safe_int(conn.params["page"]) || 1, package_count)
-    packages      = fetch_packages(page, packages_per_page, search)
+    packages      = fetch_packages(page, packages_per_page, search, sort)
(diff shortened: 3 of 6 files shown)


parkerl:
@ericmj I just took a peek at the issue and realized you wanted it to default to being sorted by downloads. My bad. How does it look other than that?

lib/hex_web/package.ex
```diff
@@ -202,6 +203,20 @@ defmodule HexWeb.Package do
   defp like_escape(string, escape) do
     String.replace(string, escape, "\\\\\\1")
   end
+
+  defp sort(query, nil) do
+    sort(query, :downloads)
+  end
```
ericmj: Looks like we don't need this clause?

lib/hex_web/package.ex
```diff
@@ -135,9 +135,10 @@ defmodule HexWeb.Package do
     |> HexWeb.Repo.delete_all
   end
 
-  def all(page, count, search \\ nil) do
+  def all(page, count, search \\ nil, sort \\ :downloads) do
```
ericmj: The main packages should sort by name, only searches should sort by downloads.

lib/hex_web/package.ex
```diff
@@ -202,6 +203,20 @@ defmodule HexWeb.Package do
...
+  end
+
+  defp sort(query, :name) do
+    from d in query, order_by: :name
+  end
+
+  defp sort(query, :downloads) do
+    from p in query,
+    left_join: d in HexWeb.Stats.PackageDownload, on: p.id == d.package_id and d.view == "all",
+    order_by: [asc: is_nil(d.downloads), desc: d.downloads]
```
ericmj: Beautiful :)

lib/hex_web/util.ex
```diff
@@ -20,6 +20,12 @@ defmodule HexWeb.Util do
     :calendar.gregorian_days_to_date(today_days - 1)
   end
 
+  def last_year do
+    {today, _time} = :calendar.universal_time()
+    today_days = :calendar.date_to_gregorian_days(today)
+    :calendar.gregorian_days_to_date(today_days - 35)
+  end
```
ericmj: This is only 35 days back, so it shouldn't be called last_year. It's only used in the sample_data script so move it there.

lib/hex_web/package.ex
ericmj: I see from the code below that we actually need this clause. Change the code below so that `nil => :name`.

ericmj:
Looks good, left a few comments and I'm going to try it out locally.

parkerl:
Changes made

lib/hex_web/web/router.ex
```diff
@@ -186,6 +187,18 @@ defmodule HexWeb.Web.Router do
     |> String.strip
   end
 
+  defp safe_sort(nil), do: nil
```
ericmj: This clause is not needed if you have the catch-all below.

lib/hex_web/package.ex
```diff
@@ -135,9 +135,10 @@ defmodule HexWeb.Package do
     |> HexWeb.Repo.delete_all
   end
 
-  def all(page, count, search \\ nil) do
+  def all(page, count, search \\ nil, sort \\ :name) do
     from(p in HexWeb.Package,
-         order_by: p.name)
+         preload: :downloads)
+    |> sort(sort || :name)
```
ericmj: Don't default here.

lib/hex_web/web/router.ex
```diff
@@ -100,14 +100,15 @@ defmodule HexWeb.Web.Router do
     conn              = fetch_params(conn)
     packages_per_page = @packages
     search            = conn.params["search"] |> safe_query
+    sort              = conn.params["sort"] |> safe_sort
```
ericmj: This is where you should default. Sort on :name by default and on :downloads if we are doing a search.

jcspencer:
This looks good to me :+1:

The only question I have is whether packages should be sorted by a drop-down _sort by_ box instead of clicking the table headers? I personallythink it would look cleaner.

lib/hex_web/web/router.ex
parkerl: Ah of course. Sorry I was in too much of a hurry this afternoon. Thanks for your patience.

ericmj:
@parkerl Great work. Please squash into a single commit and I will merge after.

@jcspencer Not sure where it would fit, feel free to play around it with after we merge.

parkerl:
Done.

jcspencer:
@ericmj I've put this together locally. What do you think?
If you're interested, I can open a PR.

screen shot 2015-05-16 at 3 16 35 pm
screen shot 2015-05-16 at 3 16 45 pm

ericmj:
Beautiful work as usual James, please open a PR after the current one is
merged, should be by the end of the day.

PR #147, opened 2015-08-23
Add user page that shows user's packages

ma2gedev:
There are many awesome packages in hex.pm.
I would like to know creators who created awesome packages like rubygems.org profile page (e.g. https://rubygems.org/profiles/josevalim). :smiley:
This pull request adds creator specific page that shows user's packages.

## main changes
- user page uri is `/users/:id` for example `/users/100`
- add user page link into `/packages/:name` page

## after this pull request

- package page shows user page links

- user page shows user's package

### DIFF
--- a/lib/hex_web/user.ex
+++ b/lib/hex_web/user.ex
@@ -137,6 +137,13 @@ defmodule HexWeb.User do
     |> HexWeb.Repo.one
   end
 
+  def get(id: id) do
+    from(u in HexWeb.User,
+         where: u.id == ^id,
+         limit: 1)
+    |> HexWeb.Repo.one
+  end
+
   def delete(user) do
     HexWeb.Repo.delete(user)
   end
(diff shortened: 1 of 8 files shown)


josevalim:
Nice!

ericmj:
Thanks @ma2gedev. I appreciate the hard work and Hex needs all help it can get but I don't think we can accept this PR unfortuntely.

The only reason we have users at all in Hex is because it's the easiest to make sure only the package owner(s) can publish the package. We never show who owns the package because that is private information, if the maintainers want they can list this information themselves on the package page either through links (to the github repo for example) or by listing themselves in the contributors field.

The API also hides the information: /api/users/:email_or_username returns 404 for everyone but the currently authed user so you can see we already hide this information.

teamon:
@ericmj Can you elaborate a bit more about the reasoning? I'm not saying it's bad, I'm just curious.

josevalim:
@teamon the reasoning is clear above: all the user information we have on the website today is private. There are no plans to make it public. The links section is a great opportunity to expand on that if someone wants to.

ma2gedev:
Thanks everyone to explain background about user information.
I understood the reason why this PR is not accepted. I'll close it.

Also thanks for all your work!

teamon:
Well, right, actually makes sense, thanks! :)

PR #153, opened 2015-09-14
add rebar3 publish and usage pages

tsloughter:
I don't like the look of the drop down but couldn't get anything I was happy with to work. So I'm opening this PR in hopes of some feedback and help styling the drop down and review of the added content.


ericmj:
Will check this out locally later today.

ericmj:
What do you think of this? Increased the font size of headers to 14px (same as normal dropdown items), darkened the color slightly and added another separator?

screen shot 2015-09-15 at 01 53 58

tsloughter:
Yea, that looks good.

PR #159, opened 2015-09-28
Sign registry with Elixir key

jcspencer:
This PR updates HexWeb.RegistryBuilder to sign the registry file on publish if the `HEX_SIGNING_KEY` environment variable is populated with a private key.

If enabled, the signature is written to `registry.ets.gz.signed` in the store.

(If this is merged, I will open a pull request to the hex client repo to support signature verification)


config/config.exs
```diff
@@ -24,7 +25,10 @@ config :hex_web,
   ses_user:      System.get_env("HEX_SES_USERNAME"),
   ses_pass:      System.get_env("HEX_SES_PASSWORD"),
 
-  secret:        System.get_env("HEX_SECRET")
+  secret:        System.get_env("HEX_SECRET"),
+
+  signing:       signing,
```
ericmj: Lets keep the config as lean as possible and just use `signing_key`. We can check for `nil` on that value.

lib/hex_web/registry_builder.ex
```diff
@@ -106,7 +107,17 @@ defmodule HexWeb.RegistryBuilder do
...
+
+    store = Application.get_env(:hex_web, :store)
+    store.put_registry(output)
+
+    if Application.get_env(:hex_web, :signing) do
+      key = Application.get_env(:hex_web, :signing_key)
+      signature = Util.sign(output, key)
+      store.put_registry_signature(signature)
+    end
+
```
ericmj: We should compute the signature from the uncompressed registry and do it on a sha512 checksum.

lib/hex_web/util.ex
```diff
@@ -224,4 +224,12 @@ defmodule HexWeb.Util do
       end)
     end
   end
+
+  def sign(file, key) do
+    [entry | _ ] = :public_key.pem_decode(key)
+    key = :public_key.pem_entry_decode(entry)
+
+    :public_key.sign(file, :sha512, key)
+    |> Base.encode64
```
ericmj: Let's use hex16 here.

ericmj:
Looks great, dropped a few comments, with some tests it's perfect.

config/config.exs
jcspencer: Agreed

lib/hex_web/registry_builder.ex
jcspencer: So are you suggesting doing something like this?:

```elixir
checksum = :crypto.hash(:sha256, output) |> Base.encode16
signature = Util.sign(output, key)
```

jcspencer: @ericmj ping

ericmj: Yes, but we shouldn't have to encode before signing.

jcspencer:
@ericmj All done.

lib/hex_web/store/local.ex
```diff
@@ -31,10 +31,19 @@ defmodule HexWeb.Store.Local do
     File.write!(Path.join(dir, "registry.ets.gz"), :zlib.gzip(data))
   end
 
+  def put_registry_signature(signature) do
+    File.mkdir_p!(dir)
+    File.write!(Path.join(dir, "registry.ets.signed"), signature)
```
ericmj: `"registry.ets.gz.signed"`

lib/hex_web/store/s3.ex
```diff
@@ -21,10 +21,18 @@ defmodule HexWeb.Store.S3 do
     upload(:s3_bucket, "registry.ets.gz", :zlib.gzip(data))
   end
 
+  def put_registry_signature(signature) do
+    upload(:s3_bucket, "registry.ets.signed", signature)
```
ericmj: `"registry.ets.gz.signed"`

ericmj:
Added two comments. Can you add tests that checks the correctness of the store? There are router integrations tests that already test it that you can look at.

lib/hex_web/store/local.ex
jcspencer: This doesn't make sense, as we're not signing the tarball, but rather the uncompressed ets registry?

ericmj: But we don't provide the ets file so it may be confusing otherwise. Also, `send_registry_signature` uses `"registry.ets.gz.signed"` so it needs to be consistent. If we had tests they would break.

jcspencer:
I agree with your two comments.

What do we need to test, consistency wise? The consistency of the signature vs. the registry?

ericmj:
Yes and that we call the store to push the registry and the signed key.

ericmj:
Ping?

jcspencer:
Sorry, I completely forgot about this pull request. I've just pushed up two unit tests.

-- 
James Spencer
Sent with Airmail

On 17 November 2015 at 10:13:22 pm, Eric Meadows-Jönsson (notifications@github.com) wrote:

Ping?

test/hex_web/registry_builder_test.exs
```diff
@@ -72,6 +76,63 @@ defmodule HexWeb.RegistryBuilderTest do
...
+    tmp = Application.get_env(:hex_web, :tmp)
+    reg = File.read!(Path.join(tmp, "store/registry.ets.gz")) |> :zlib.gunzip
+    sig = File.read!(Path.join(tmp, "store/registry.ets.gz.signed"))
+
+    checksum = :crypto.hash(:sha512, reg)
+
+    assert HexWeb.Util.sign(checksum, key) == sig
+  end
+
+  test "integration fetch registry" do
```
ericmj: Tag this test with `:integration`

ericmj:
I am now thinking that maybe `registry.ets.signed` is the correct name of the signature file. Since I've been going back and forth I will change it myself instead :).

Thanks for this James, will you be working on the client also?

jcspencer:
Yeah, I have a working prototype of the client. I am currently overseas, but I can complete it in early January

On 8 December 2015 6:47:44 PM GMT+07:00, "Eric Meadows-Jönsson" <notifications@github.com> wrote:
>I am now thinking that maybe `registry.ets.signed` is the correct name
>of the signature file. Since I've been going back and forth I will
>change it myself instead :).
>
>Thanks for this James, will you be working on the client also?
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/hexpm/hex_web/pull/159#issuecomment-162859392

-- 
James Spencer

ericmj:
In that case I will work on it :). Feel free to share the prototype code if you'd like.

jcspencer:
Can the functionality of this be enabled server-side?

ericmj:
Sorry for the delay on this, it's on my todo list.

jcspencer:
I'm currently working on more unit tests for the server side, and then it should be ready to push into production. Currently, we're unnecessarily double hashing the digest.

PR #160, opened 2015-10-02
Enable syntax highlighting for doc code blocks

jcspencer:
This PR uses [highlight.js](https://highlightjs.org) to highlight the Elixir and Erlang syntax in the documentation.


jcspencer:
Demo:
demo

ericmj:
Beautiful! Thank you.

ericmj:
Do you know if the highlight error of `defp` is reported to highlight.js?

jcspencer:
Not that I know of.. that's a strange issue. I'll look into it.

jcspencer:
The issue looks to be that the elixir highlighting code is not included in the default highlight.js bundle, and is defaulting to Ruby.

I've pushed a fix that includes a custom highlight.js bundle with *only* erlang and elixir support.

PR #186, opened 2015-11-19
Add unversioned hexdocs link to package view

fteem:
Related to: https://github.com/hexpm/hex_web/issues/169

The code checks if any of the package releases have docs (which, if I am not mistaken, confirms that there's a hexdocs.pm page for the package) and renders the link in the template if the docs are present.

Thoughts?

### DIFF
--- a/lib/hex_web/web/templates/package.html.eex
+++ b/lib/hex_web/web/templates/package.html.eex
@@ -50,6 +50,14 @@ build_tools = @current_release && @current_release.meta["build_tools"] || []
           </ul>
         </dd>
       <% end %>
+
+      <%= if @hexdocs_url != nil and @docs_tarball_url != nil do %>
+        <dt>Docs</dt>
+        <dd>
+          <a href="<%= @hexdocs_url %>">online</a>
+          (<a href="<%= @docs_tarball_url %>">download</a>)
+        </dd>
+      <% end %>
     </dl>
     <div style="margin: 10px 0; border-top: 1px solid #eee"></div>
 
--- a/lib/hex_web/util.ex
+++ b/lib/hex_web/util.ex
@@ -141,6 +141,14 @@ defmodule HexWeb.Util do
   end
 
   @doc """
+  Returns a url to the documentation tarball in the Amazon S3 Hex.pm bucket.
+  """
+  @spec docs_tarball_url(String.t) :: String.t
+  def docs_tarball_url(package_name) do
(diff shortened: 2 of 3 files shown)


lib/hex_web/web/templates/package.html.eex
```diff
@@ -50,6 +50,11 @@ build_tools = @current_release && @current_release.meta["build_tools"] || []
           </ul>
         </dd>
       <% end %>
+
+      <%= if @hexdocs_url do %>
+        <dt>Hexdocs</dt>
```
ericmj: Just call this "Documentation" instead.

ericmj: Hum, actually hold on, just saw the line below. I think we may need another place for this.

fteem: The conditional logic in the router seems to be out of place, but I am not
sure what would be the best place to put it.

ericmj: The conditional is fine, I think we need another place for the link on the page. I don't think "Documentation: Hexdocs" will look good.

fteem:
@ericmj As a side question - is there a way to make the server recompile the templates without restarting the server?

ericmj:
`r ModuleThatDefinesTemplates`. It will be automatic when we move to phoenix.

lib/hex_web/web/templates/package.html.eex
fteem: take1
take2

Just ideas...

fteem: Another idea :)
take3

ericmj: Sorry for the back and forth, I talked to José and we found a design that we like. Lets go with your original approach, but with a slight difference: "Docs: online (download)".

"online" will have the hexdocs link. "download" will link to the tarballed documentation on S3. An example URL would be https://s3.amazonaws.com/s3.hex.pm/docs/decimal.tar.gz. You can build it with the `s3_url` and `s3_bucket` configuration values. The link will be broken at first because I need to setup redirects from decimal.tar.gz to decimal-LATEST.tar.gz.

https://github.com/hexpm/hex_web/blob/master/config/config.exs#L17-L18

fteem:
@ericmj I updated the UI with your suggestion, check it out. Here's a screeshot as well: https://monosnap.com/file/iaVrlwbuh1wzT053L0uiq2c5tOdQpJ

ericmj:
Looks great, please squash into a single commit. I will merge after setting up the tarball links.

PR #225, opened 2016-02-02
Frontend Proposal [WIP]

pragmaticivan:
I've been working on migrating the new template and design. I'm opening this PR now so you guys can follow my progress.

1st proposal (external project): http://pragmaticivan.github.io/hex-web-template

If possible, I would love to have a dump file of packages so I can host in my local database and test with consistent data.

cc @ericmj


lexmag:
Several files have changed their permissions 100644 → 100755.

package.json
```diff
@@ -1,12 +1,12 @@
...
-  },
+  "repository": {},
   "dependencies": {
-    "brunch": "^1.8.5",
     "babel-brunch": "^5.1.1",
+    "brunch": "^1.8.5",
     "clean-css-brunch": ">= 1.0 < 1.8",
     "css-brunch": ">= 1.0 < 1.8",
     "javascript-brunch": ">= 1.0 < 1.8",
+    "sass-brunch": "^1.9.1",
```
lexmag: Though I'm not an expert in frontend technologies, I'm wondering if Sass usage will make it harder to maintain for people with no Sass experience.
Also I've seen there is https://github.com/postcss/postcss available, which seems to provide Sass features but doesn't introduce new syntax.
It's the thing behind well-known https://github.com/postcss/autoprefixer.
:bowtie:

web/templates/docs/publish.html.eex
```diff
@@ -1,5 +1,7 @@
-<h2>Publishing a package</h2>
-
+<div class="container page">
+<h2 class="page-title">
+  Publishing a package
+</h2>
```
ericmj: Looks like `page-title` is not needed?

web/templates/docs/usage.html.eex
```diff
@@ -1,24 +1,34 @@
-<h2>Usage</h2>
+<div class="container page">
+      <h2 class="page-title">
```
ericmj: Too much indentation.

web/templates/layout/header.html.eex
```diff
@@ -5,31 +5,37 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta name="description" content="A package manager for the Erlang ecosystem">
-    <link rel="search" type="application/opensearchdescription+xml" title="Hex" href="/hexsearch.xml">
-    <link rel="stylesheet" href="<%= static_path(HexWeb.Endpoint, "/css/app.css") %>">
+
     <title><%= title(assigns) %></title>
+
+    <link rel="search" type="application/opensearchdescription+xml" title="Hex" href="/hexsearch.xml">
+    <link href='https://fonts.googleapis.com/css?family=Lato:400,300,700,900,100' rel='stylesheet' type='text/css'>
```
ericmj: We should vendor this if possible.

web/templates/layout/header.html.eex
```diff
@@ -5,31 +5,37 @@
...
+    <link rel="stylesheet" href="<%= static_path(HexWeb.Endpoint, "/css/main.css") %>">
     <link rel="shortcut icon" href="/favicon.png">
+
   </head>
   <body>
-    <nav class="navbar navbar-default navbar-static-top" role="navigation">
+    <!--[if lt IE 10]>
+      <p class="browserupgrade">You are using an <strong>outdated</strong> browser. Please <a href="http://browsehappy.com/">upgrade your browser</a> to improve your experience.</p>
+    <![endif]-->
```
ericmj: What do we need from IE10?

ericmj:
This is awesome @pragmaticivan, we greatly appreciate all the work you are putting into this.

A few small things: Can we use the original hex logo instead of the one modified with gray lines. Also the favicon is using the old logo, we should be able to use the existing ones I think? Can we move the vendored css and js to the vendor directories.

pragmaticivan:
@ericmj Sure, I just used the one I had in my template, but I have plans to change the logo and favicon for the new one, I just need you to send me in private the final version. As I said, It's just a WIP, I hope to improve a bunch of things.

package.json
pragmaticivan: Lex It would make harder if I use the `.sass` format. But in this project, I'm using `.scss` which try to simulates the normal way we use for css development. You can take a look at the new assets.

pragmaticivan: Please check this file as an example: https://github.com/pragmaticivan/hex_web/blob/feature/new-frontend/web/static/css/template/_footer.scss

web/templates/docs/publish.html.eex
pragmaticivan: Not anymore, and I'm also moving the main template to the main layout, but I need to setup a variable for that in  templates we need to change and add an aditional class.

web/templates/layout/header.html.eex
pragmaticivan: Yes, I can add in main.scss but not clear if it's better if we use like that, because it's using a CDN and will definitely load faster if we use google version. Haven't tested also if it will break with sass. I'll play with that today.

web/templates/layout/header.html.eex
pragmaticivan: It's just a common way we use to tell people to upgrade their browsers in case they're using internet explorer. It never shows in the front-end unless they have IF below 10. I can remove, but it's a good way to tell people to upgrade their browser.

ericmj:
Just let me know what sizes of the logo you need and I'll send them over.

web/templates/layout/header.html.eex
pragmaticivan: Here's the project I was talking about: http://browsehappy.com/

pragmaticivan:
@ericmj I just updated the logo and seems not "normal" with the header, thats why I created the little shadow in the logo before. Any way, I just left there with the old one. Probably the favicon problem is cache. I'm using exactly the one you have in master.

pragmaticivan:
What do you think about those links?
links

I just added this area in case ou guys want to have a featured links area. And also if you're thinking about having a new area there. Maybe a Contribution Guide or a Post a Job area.

ericmj:
The favicon seem to have been fixed in the latest commit. I don't think we need those links right now, we don't really have anything to put there except for installation guides which we already at the top of the page.

Also, if you rebase against master you will get rid of the extra commits.

pragmaticivan:
Hey @ericmj I just rebased from master but I got a problem with a helper here, the variable coming for the controller is nil, I don't know what's happening, I'm not very familiar with phoenix yet. But got some free time and Would like to finish this PR asap. Can you help with this error?

ericmj:
Have you checked that the values are set correctly in the controller? https://github.com/pragmaticivan/hex_web/blob/feature/new-frontend/web/controllers/page_controller.ex#L17-L19

My guess is that there are no downloads in the database which could cause this behaviour.

mitchellhenke:
@ericmj could the change from string keys to atom keys be causing the issue?

https://github.com/hexpm/hex_web/pull/225/files#diff-35eea261dbc7630db064608d98650223L38 vs.
https://github.com/hexpm/hex_web/pull/225/files#diff-35eea261dbc7630db064608d98650223R87

ericmj:
@mitchellhenke That's probably it, nice find.

pragmaticivan:
Thanks guys, the `"` was the problem, but was a good one, now we would have a necessity of having `human_number_space` expecting nil :)

ericmj:
Although if `human_number_space` accepted `nil` we wouldn't have gotten an error. It would just silently return the wrong result.

pragmaticivan:
@ericmj can you add a label here for `help wanted`?

ericmj:
@pragmaticivan What is it you need help with?

pragmaticivan:
@ericmj there are some new queries and adjustments for back-end. My proposal was to do the front-end. I still learning Elixir and didn't want to jump fast in Phoenix without get a better knowledge, but that's ok, once I finish my studies I can do the back-end changes for that. The front-end is almost done for this PR.

ericmj:
I will work on the rebase to make this mergable with master again. Then we can take a look at what's left to do on the frontend. Thank you for everything you've done so far @pragmaticivan! :heart:

ericmj:
@pragmaticivan Sent you a PR with the master merge. Let's see if that will make it mergable into master.

https://github.com/pragmaticivan/hex_web/pull/1

ericmj:
@pragmaticivan Are you leaving this PR for the community to continue work on? If so, would you mind if recreate the PR with a rebased branch?

pragmaticivan:
@ericmj Go for, would be even better for me if you have the branch inside the repo, so I can create PRs to merge inside this branch.

ericmj:
Will do, thanks for all you work :heart:

pragmaticivan:
I think you can close this PR since it has a new branch.

ericmj:
@wojtekmach Is it okay if I move your rebased branch over to this repo so people can continue to do PRs to it?

wojtekmach:
@ericmj of course, please go ahead. cc @pragmaticivan

ericmj:
Added it here: https://github.com/hexpm/hex_web/tree/frontend

PR #238, opened 2016-02-15
Add audit log

wojtekmach:
See: https://github.com/hexpm/hex_web/issues/202

### DIFF
--- a/web/controllers/api/key_controller.ex
+++ b/web/controllers/api/key_controller.ex
@@ -23,7 +23,18 @@ defmodule HexWeb.API.KeyController do
   end
 
   def create(conn, params) do
-    case Key.create(conn.assigns.user, conn.params) |> HexWeb.Repo.insert do
+    result =
+      HexWeb.Repo.transaction(fn ->
+        case Key.create(conn.assigns.user, conn.params) |> HexWeb.Repo.insert do
+          {:ok, key} ->
+            audit(conn, "key.generate", key)
+            key
+          {:error, changeset} ->
+            HexWeb.Repo.rollback(changeset)
+        end
+      end)
+
+    case result do
       {:ok, key} ->
         location = key_url(conn, :show, params["name"])
 
@@ -32,14 +43,17 @@ defmodule HexWeb.API.KeyController do
         |> api_cache(:private)
         |> put_status(201)
         |> render(:show, key: key)
-      {:error, errors} ->
-        validation_failed(conn, errors)
+      {:error, changeset} ->
+        validation_failed(conn, changeset.errors)
     end
   end
 
   def delete(conn, %{"name" => name}) do
     if key = HexWeb.Repo.one(Key.get(name, conn.assigns.user)) do
-      HexWeb.Repo.delete!(key)
+      HexWeb.Repo.transaction(fn ->
+        HexWeb.Repo.delete!(key)
+        audit(conn, "key.remove", key)
+      end)
 
       conn
       |> api_cache(:private)
--- a/web/controllers/api/owner_controller.ex
+++ b/web/controllers/api/owner_controller.ex
@@ -29,7 +29,18 @@ defmodule HexWeb.API.OwnerController do
     user = HexWeb.Repo.get_by!(User, email: email)
     package = conn.assigns.package
 
-    case Package.create_owner(conn.assigns.package, user) |> HexWeb.Repo.insert do
+    result =
+      HexWeb.Repo.transaction(fn ->
+        case Package.create_owner(conn.assigns.package, user) |> HexWeb.Repo.insert do
+          {:ok, owner} ->
+            audit(conn, "owner.add", {package, user})
+            owner
+          {:error, changeset} ->
+            HexWeb.Repo.rollback(changeset)
+        end
+      end)
+
+    case result do
       {:ok, _} ->
         owners = assoc(package, :owners) |> HexWeb.Repo.all
 
@@ -60,8 +71,10 @@ defmodule HexWeb.API.OwnerController do
       |> api_cache(:private)
       |> send_resp(403, "")
     else
-      Package.owner(package, owner)
-      |> HexWeb.Repo.delete_all
+      HexWeb.Repo.transaction(fn ->
+        Package.owner(package, owner) |> HexWeb.Repo.delete_all
+        audit(conn, "owner.remove", {package, owner})
+      end)
 
       HexWeb.Mailer.send(
         "owner_remove.html",
--- a/web/controllers/controller_helpers.ex
+++ b/web/controllers/controller_helpers.ex
@@ -163,4 +163,17 @@ defmodule HexWeb.ControllerHelpers do
     fun = Keyword.get(opts, :fun, fn _, _ -> true end)
     HexWeb.AuthHelpers.authorized(conn, opts, &fun.(conn, &1))
   end
+
+  @doc """
+  Records an entry in audit log. This function should be used within the same transaction
+  as DB operations that are part of the action that is being audited.
+  """
+  def audit(%Plug.Conn{assigns: %{user: user}}, action, params) do
+    audit(user, action, params)
+  end
+
+  def audit(user, action, opts) do
+    HexWeb.AuditLog.create(user, action, opts)
+    |> HexWeb.Repo.insert!
+  end
 end
--- a/web/models/audit_log.ex
+++ b/web/models/audit_log.ex
@@ -1,0 +1,31 @@
+defmodule HexWeb.AuditLog do
+  use HexWeb.Web, :model
+
+  schema "audit_logs" do
+    belongs_to :actor, User
+    field :action, :string
+    field :params, :map
+    timestamps(updated_at: false)
+  end
+
+  def create(%HexWeb.User{id: user_id}, action, params) do
+    params = extract_params(action, params)
+    %HexWeb.AuditLog{actor_id: user_id, action: action, params: params}
+  end
+
+  defp extract_params("docs.publish", {package, release}), do: %{package: serialize(package), release: serialize(release)}
+  defp extract_params("docs.revert", {package, release}), do: %{package: serialize(package), release: serialize(release)}
+  defp extract_params("key.generate", key), do: serialize(key)
+  defp extract_params("key.remove", key), do: serialize(key)
+  defp extract_params("owner.add", {package, user}), do: %{package: serialize(package), user: serialize(user)}
+  defp extract_params("owner.remove", {package, user}), do: %{package: serialize(package), user: serialize(user)}
+  defp extract_params("release.publish", {package, release}), do: %{package: serialize(package), release: serialize(release)}
+  defp extract_params("release.revert", {package, release}), do: %{package: serialize(package), release: serialize(release)}
+
+  defp serialize(schema), do: Map.take(schema, fields(schema))
+
+  defp fields(%HexWeb.Key{}), do: [:id, :name]
+  defp fields(%HexWeb.Package{}), do: [:id, :name, :meta]
+  defp fields(%HexWeb.Release{}), do: [:id, :version, :checksum, :meta, :has_docs, :package_id]
+  defp fields(%HexWeb.User{}), do: [:id, :username, :email, :confirmed]
+end
--- a/priv/repo/migrations/20160215102451_add_audit_logs_table.exs
+++ b/priv/repo/migrations/20160215102451_add_audit_logs_table.exs
@@ -1,0 +1,12 @@
+defmodule HexWeb.Repo.Migrations.AddAuditLogsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:audit_logs) do
+      add :actor_id, references(:users)
+      add :action, :string, null: false
+      add :params, :jsonb, null: false
+      add :inserted_at, :timestamp, null: false
+    end
+  end
+end
--- a/test/controllers/api/docs_controller_test.exs
+++ b/test/controllers/api/docs_controller_test.exs
@@ -26,6 +26,11 @@ defmodule HexWeb.API.DocsControllerTest do
     assert conn.status == 201
     assert HexWeb.Repo.get_by!(assoc(phoenix, :releases), version: "0.0.1").has_docs
 
+    log = HexWeb.Repo.one!(HexWeb.AuditLog)
+    assert log.actor_id == user.id
+    assert log.action == "docs.publish"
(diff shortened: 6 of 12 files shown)


wojtekmach:
If this initial implementation looks good, I'll add auditing for the remaining controllers mentioned in the ticket

web/models/audit_log.ex
```diff
@@ -0,0 +1,51 @@
...
+
+  schema "audit_logs" do
+    field :action, :string
+    field :extra_params, :map
+    timestamps(updated_at: false)
+  end
+
+  def create!(conn, action, opts) do
+    build(conn, action, opts)
+    |> HexWeb.Repo.insert!
```
ericmj: Don't access the repo in the model.

web/models/audit_log.ex
```diff
@@ -0,0 +1,51 @@
...
+  defp extract_extra_params("release.revert", conn, %HexWeb.Release{id: id}) do
+    with_conn(conn, %{release_id: id})
+  end
+  defp extract_extra_params("user.reset_password", _conn, %HexWeb.User{id: id}) do
+    %{user_id: id}
+  end
+
+  defp with_conn(%Plug.Conn{assigns: %{user: user}}, params) do
+    Map.merge(%{actor_id: user.id}, params)
+  end
```
ericmj: I think it's nicer to pass the user in explicitly then have the model depend on plug.

web/models/audit_log.ex
```diff
@@ -0,0 +1,51 @@
+defmodule HexWeb.AuditLog do
+  use HexWeb.Web, :model
+
+  schema "audit_logs" do
+    field :action, :string
+    field :extra_params, :map
```
ericmj: There is always a specific user performing an action so I think the user should be stored in a field rather than in extra_params.

ericmj:
We need to make sure that the audit log creation happens in a transaction with the same action that is being logged.

web/models/audit_log.ex
```diff
@@ -0,0 +1,51 @@
...
+  end
+  defp extract_extra_params("release.publish", conn, %HexWeb.Release{id: id}) do
+    with_conn(conn, %{release_id: id})
+  end
+  defp extract_extra_params("release.revert", conn, %HexWeb.Release{id: id}) do
+    with_conn(conn, %{release_id: id})
+  end
+  defp extract_extra_params("user.reset_password", _conn, %HexWeb.User{id: id}) do
+    %{user_id: id}
+  end
```
ericmj: I don't think reset_password needs to be logged since we have no information about who initiated it.

web/models/audit_log.ex
wojtekmach: makes sense

web/models/audit_log.ex
wojtekmach: @ericmj yeah, not logging this action would simplify the implementation a bit. However, reset_password was mentioned in the original issue and e.g. github does that too. Again, if we can keep this out of the scope, the implementation should be simpler and schema more consistent

web/models/audit_log.ex
wojtekmach: Souds good

web/controllers/api/owner_controller.ex
```diff
@@ -27,8 +27,10 @@ defmodule HexWeb.API.OwnerController do
   def create(conn, %{"email" => email}) do
     email = URI.decode_www_form(email)
     user = HexWeb.Repo.get_by!(User, email: email)
+    package = conn.assigns.package
 
-    Package.create_owner(conn.assigns.package, user) |> HexWeb.Repo.insert!
+    Package.create_owner(package, user) |> HexWeb.Repo.insert!
+    HexWeb.AuditLog.create!(conn, "owner.add", {package, user})
```
wojtekmach: @ericmj regarding your 2 comments about not accessing Repo in Models, and passing in user explicitly. I think the usage in controllers would then be something along the lines of:

```elixir
HexWeb.Repo.transaction(fn ->
  Package.create_owner(package, user) |> HexWeb.Repo.insert!
  HexWeb.AuditLog.build(conn.assigns.user, "owner.add", {package, user}) |> HexWeb.Repo.insert!
end)
```

The audit invocation is pretty verbose and its gonna be in almost every controller, so I'm thinking about potentially moving some boilerplate to a helper, but will have a better idea if it's worth it when I update all controllers first with transactions

ericmj: That looks fine. When Ecto.Multi ships we will be able to clean it up a bit.

web/controllers/api/key_controller.ex
```diff
@@ -25,6 +25,7 @@ defmodule HexWeb.API.KeyController do
   def create(conn, params) do
     case Key.create(conn.assigns.user, conn.params) |> HexWeb.Repo.insert do
       {:ok, key} ->
+        HexWeb.AuditLog.create(conn.assigns.user, "key.generate", key) |> HexWeb.Repo.insert!
```
wojtekmach: @ericmj I wrapped most of the DB operations and audits within the same transaction but here I encountered an interesting problem. We pattern match the return value of Repo.insert to gracefully handle validation errors, but when this operation would be wrapped in a transaction, `Repo.transaction(fun)` would either return `{:ok, result_of_fun}` or `{:error, :rollback}` in which keys the changeset.errors are somewhat lost. Can you think of any elegant solution? I tried stuff like this: https://gist.github.com/wojtekmach/84a81416b39815de93f3, but it led me nowhere.

Similar problem here: https://github.com/wojtekmach/hex_web/blob/audit_logs/web/controllers/api/release_controller.ex#L112:L114

ericmj: You need to explicitly call Repo.rollback with the changeset.

ericmj: The validation_failed looks wrong indeed.

wojtekmach: thanks for the tip, will look into Repo.rollback

jcspencer:
Looks good to me (besides Eric’s comments).

The only possible issue I can see is that there are some cases where a task may fail but the audit log is still appended to. Do we need to ensure that tasks run as expected before logging them? What are your thoughts?

web/controllers/api/key_controller.ex
```diff
@@ -23,7 +23,18 @@ defmodule HexWeb.API.KeyController do
   end
 
   def create(conn, params) do
-    case Key.create(conn.assigns.user, conn.params) |> HexWeb.Repo.insert do
+    result =
+      HexWeb.Repo.transaction(fn ->
+        case Key.create(conn.assigns.user, conn.params) |> HexWeb.Repo.insert do
+          {:ok, key} ->
+            HexWeb.AuditLog.create(conn.assigns.user, "key.generate", key) |> HexWeb.Repo.insert!
```
wojtekmach: @ericmj I pushed the changes with rollback as we discussed. Do you think it's OK to do Repo.insert! for audit log, or it would be better to handle possible error here as well? I don't really expect this to ever fail other than when the DB is offline or full.

cc @jcspencer

ericmj: Repo.insert! is perfect.

wojtekmach: :+1:

I'm considering doing one more addition which would clean up the controllers a little bit, basically:

```elixir
# controller_helpers.ex

# decent docs explaining this should be running in a transaction
def audit(conn, action, opts) do
 ...
end

# key_controller.ex
audit("key.generate", key)
```

what do you think?

ericmj: Perfect. This looks ready to go after some testing is added.

wojtekmach:
@ericmj I updated the PR with `audit` helper function. You mentioned something about missing test coverage, did you have something particular in mind? I could add something like: `assert HexWeb.Repo.all(HexWeb.AuditLog) |> length == 1` for every controller action that is audited (I didn't add it initially because it seems somewhat redundant)

ericmj:
Tests for the model and a check that the audit log is correct (more than just checking if it was inserted) after each controller action test.

wojtekmach:
@ericmj I added missing controller tests (and found a bug), so thanks for pointing that out. IMHO the one model test that I originally had should be enough because adding more (which would test the format of log) would be redundant with the controller tests

ericmj:
The only step left is to make sure we never delete the things the audit log references. This is tricky because it either means adding a `field :deleted, :boolean` with index on all those models and always filtering on deleted or adding a separate table where we move deleted things.

Another idea would be to store a snapshot of the model in the audit log. On second thought this might actually be the best idea.

wojtekmach:
@ericmj

> Another idea would be to store a snapshot of the model in the audit log. On second thought this might actually be the best idea.

Agreed, I think this would be good. This would also provide more information (e.g. in future cases where things could be e.g. later renamed it's good to log the values "at the time" of change). Any specific fields we should take as part of the snapshot, or just the entire model (along with timestamps etc)? I'll make the appropriate changes in next couple of days and rebase.

> The only step left is to make sure we never delete the things the audit log references.

I think even if we go with snapshot in the audit log, doing "soft deletes" on most/all entities could still be useful in certain cases (e.g. you want to quickly undelete something by just swapping the flag, you want to see the number of e.g. packages ever created etc) but perhaps it could then be outside of the scope of this PR which is already a bit long.

ericmj:
> Agreed, I think this would be good. This would also provide more information (e.g. in future cases where things could be e.g. later renamed it's good to log the values "at the time" of change). Any specific fields we should take as part of the snapshot, or just the entire model (along with timestamps etc)? I'll make the appropriate changes in next couple of days and rebase.

Dump the entire model I think.

> I think even if we go with snapshot in the audit log, doing "soft deletes" on most/all entities could still be useful in certain cases (e.g. you want to quickly undelete something by just swapping the flag, you want to see the number of e.g. packages ever created etc) but perhaps it could then be outside of the scope of this PR which is already a bit long.

You are right that it should be done in another PR. It's particular tricky because we have uniqueness constraints that we don't want to trigger for soft deleted entities.

web/models/audit_log.ex
```diff
@@ -0,0 +1,31 @@
...
+    field :extra_params, :map
+    timestamps(updated_at: false)
+  end
+
+  def create(%HexWeb.User{id: user_id}, action, opts) do
+    extra_params = extract_extra_params(action, opts)
+    %HexWeb.AuditLog{actor_id: user_id, action: action, extra_params: extra_params}
+  end
+
+  defp extract_extra_params("docs.publish", release), do: serialize(release)
```
wojtekmach: @ericmj I just pushed updates. Couple of notes:

- as in previous revisions, I decided to explicitly pattern match on all available actions (docs.publish, docs.revert etc) even though it's not actually needed - but it serves here as kind of live documentation of what we support in just one place
- as you can see below in `fields/1` private function, I'm explicitly whitelisting fields to audit. I thought about using reflection (__schema__/1 etc) to automatically do that, however there are fields that we probably don't want to store like users.password, keys.secret_first etc
- as you can see in controller tests, I somewhat made the assertion more relaxed where I pattern match on a subset of the map (instead of asserting the entire log format, which would make the test somewhat bloated)

ericmj:
This is great @wojtekmach. It looks ready to merge to me but I am going to ask @josevalim and @lexmag to do a final review since this has security implications. Thank you!

web/controllers/controller_helpers.ex
```diff
@@ -163,4 +163,17 @@ defmodule HexWeb.ControllerHelpers do
...
+
+  @doc """
+  Records an entry in audit log. This function should be used within the same transaction
+  as DB operations that are part of the action that is being audited.
+  """
+  def audit(%Plug.Conn{assigns: %{user: user}}, action, opts) do
+    audit(user, action, opts)
+  end
+
+  def audit(user, action, opts) do
```
lexmag: `opts` name here feels confusing to me, it'd be better to name it's `params` as it's what it represents: params for an action (maybe `args`?). :bowtie:

web/models/audit_log.ex
```diff
@@ -0,0 +1,31 @@
+defmodule HexWeb.AuditLog do
+  use HexWeb.Web, :model
+
+  schema "audit_logs" do
+    belongs_to :actor, User
+    field :action, :string
+    field :extra_params, :map
```
lexmag: Can it be just `:params`?

wojtekmach: the way I was thinking about it was: params=base_params+extra_params,base_params being actor_id, action, inserted_at. I think it's fine either way; params would be ok here with that other 'opts' variable name you mentioned elsewhere

ericmj: Agree that it can simply be `params`.

wojtekmach: Sounds good, fixed in: d73f282

2016-02-28 5:06 GMT+01:00 Eric Meadows-Jönsson <notifications@github.com>:

> In web/models/audit_log.ex
> <https://github.com/hexpm/hex_web/pull/238#discussion_r54343921>:
>
> > @@ -0,0 +1,31 @@
> > +defmodule HexWeb.AuditLog do
> > +  use HexWeb.Web, :model
> > +
> > +  schema "audit_logs" do
> > +    belongs_to :actor, User
> > +    field :action, :string
> > +    field :extra_params, :map
>
> Agree that it can simply be params.
>
> —
> <https://github.com/hexpm/hex_web/pull/238/files#r54343921>.
>



-- 
Wojtek Mach

ericmj:
After @josevalim signs off we are good to go.

ericmj:
@wojtekmach This is ready for merging. I think one of my last pushes caused a merge conflict on this, can you fix it so you get the contribution?

Thanks again for all your contributions! :heart:

wojtekmach:
Sure, no problem - I'll let you know when it's rebased! Btw, I noticed that
at least one migration was added on master since I created this PR - do I
need to do anything special (e.g. rename migration file) or Ecto would do
the right thing?

2016-03-04 19:04 GMT+01:00 Eric Meadows-Jönsson <notifications@github.com>:

> @wojtekmach <https://github.com/wojtekmach> This is ready for merging. I
> think one of my last pushes caused a merge conflict on this, can you fix it
> so you get the contribution?
>
> Thanks again for all your contributions! [image: :heart:]
>
> —
> <https://github.com/hexpm/hex_web/pull/238#issuecomment-192385348>.
>



-- 
Wojtek Mach

josevalim:
Ecto should do the correct thing. :)

wojtekmach:
@ericmj rebased! I got a build failure; it seems unrelated to this PR, but let me know if I can help with that.
@josevalim :-)

PR #263, opened 2016-03-07
Refactor `validation_failed/2` helper

wojtekmach:


wojtekmach:
By using Ecto.Changeset.traverse_errors we can remove some unnecessary code, in particular `put_embed_error` (sorry for churn :)). The code got a little bit more complicated because of the errors normalization, however it's net negative LOC and I suppose uses Ecto more as intended. WDYT?

PR #265, opened 2016-03-09
Update to Ecto 2.0.0

wojtekmach:
#See: #257

TODO:
- [x] fix flaky test: `mix test --seed 274679`. See: https://github.com/hexpm/hex_web/pull/265/commits/9dd05dcfd338c5d2b09bd18f859d852e877a9330
- [x] revert changes to PackageDownload schema - fix potential bug with association on materialized view. See: https://github.com/hexpm/hex_web/pull/265/commits/2bff1e036d14d30fd01ea485050b8fce8cfd4d9d
- [x] revert changes to Requirement related to dependency resolution - make sure we catch dependency conflicts. Might be tricky because we add one requirement at a time with cast_assoc
- [x] Resetting a user and all their keys can use either Ecto.Multi or cast_assoc.
- [x] ~~Clean up HexWeb.Release.requirements/1 if the beta supports %{model | ...} select syntax~~. It's not part of the beta
- [x] Include type error information on validation. Update to latest Ecto, and use `:type` in `traverse_errors`. See: https://github.com/elixir-lang/ecto/pull/1341
- [x] Update to latest Ecto and change usage of `Changeset.traverse_error`. See: https://github.com/elixir-lang/ecto/pull/1327
- [x] Preload empty release requirements (see comment inline). See: https://github.com/elixir-lang/ecto/issues/1305
- [x] Update to latest Ecto and use `Ecto.Multi.insert/update/delete` with `changeset_fun` for audit logs. See: https://github.com/elixir-lang/ecto/pull/1407
- [x] fix test that uses mailer (see 601cef6, #328)


wojtekmach:
@ericmj here's an early draft, still quite a few things left to do but any early feedback is much appreciated

web/controllers/api/docs_controller.ex
```diff
@@ -163,16 +163,13 @@ defmodule HexWeb.API.DocsController do
...
 
-    # Set docs flag on release
-    HexWeb.Repo.transaction(fn ->
-      Ecto.Changeset.change(release, has_docs: true)
-      |> HexWeb.Repo.update!
+    multi =
+      Ecto.Multi.new
+      |> Ecto.Multi.update(:release, Ecto.Changeset.change(release, has_docs: true))
+      |> Ecto.Multi.update(:package, Ecto.Changeset.change(release.package, docs_updated_at: Ecto.DateTime.utc))
+      |> Ecto.Multi.run(:log, fn _ -> audit(user, "docs.publish", {package, release}) end)
```
wojtekmach: Since this audit log doesn't depend on data from previous steps, it could be simply:

`|> Ecto.Multi.insert(:log, audit(user, "docs.publish", {package, release})`

however I made it that `audit/3` calls Repo.insert under the hood. Moreover, I think it would be weird to sometimes log with Multi.insert and sometimes with Multi.run. I think it would be neat to allow passing fun to Ecto.insert as proposed here: https://groups.google.com/forum/#!topic/elixir-ecto/zrJtqiR0-gU

test/fixtures/sitemap.xml
```diff
@@ -4,7 +4,7 @@
     xsi:schemaLocation="http://www.sitemaps.org/schemas/sitemap/0.9 http://www.sitemaps.org/schemas/sitemap/0.9/sitemap.xsd">
     <url>
         <loc>https://hex.pm/packages/postgrex</loc>
-        <lastmod>2014-04-17T14:00:00Z</lastmod>
+        <lastmod>2014-04-17T14:00:00</lastmod>
```
wojtekmach: need to double-check why this was needed

web/models/package_download.ex
```diff
@@ -2,11 +2,11 @@ defmodule HexWeb.PackageDownload do
   use HexWeb.Web, :model
   import HexWeb.QueryAPI, only: [coalesce: 2]
 
-  @primary_key false
```
wojtekmach: without it I was getting this:

```
48) test index (HexWeb.PageControllerTest)
     test/controllers/page_controller_test.exs:32
     ** (Ecto.NoPrimaryKeyFieldError) schema `HexWeb.PackageDownload` has no primary key
     stacktrace:
       (ecto) lib/ecto/changeset/relation.ex:355: Ecto.Changeset.Relation.primary_keys!/1
       (ecto) lib/ecto/changeset/relation.ex:125: Ecto.Changeset.Relation.change/3
       (ecto) lib/ecto/repo/schema.ex:410: anonymous fn/4 in Ecto.Repo.Schema.surface_changes/4
       (elixir) lib/enum.ex:1493: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto) lib/ecto/repo/schema.ex:398: Ecto.Repo.Schema.surface_changes/4
       (ecto) lib/ecto/repo/schema.ex:160: Ecto.Repo.Schema.do_insert/4
       (ecto) lib/ecto/repo/schema.ex:108: Ecto.Repo.Schema.insert!/4
       test/controllers/page_controller_test.exs:19: HexWeb.PageControllerTest.__ex_unit_setup_1/1
       test/controllers/page_controller_test.exs:1: HexWeb.PageControllerTest.__ex_unit__/2
```

need to investigate.

test/controllers/api/docs_controller_test.exs
```diff
@@ -13,10 +13,10 @@ defmodule HexWeb.API.DocsControllerTest do
 
   @tag :integration
   test "release docs" do
-    user           = HexWeb.Repo.get_by!(User, username: "eric")
-    {:ok, phoenix} = Package.create(user, pkg_meta(%{name: "phoenix", description: "Web framework"}))
-    {:ok, _}       = Release.create(phoenix, rel_meta(%{version: "0.0.1", app: "phoenix"}), "")
-    {:ok, _}       = Release.create(phoenix, rel_meta(%{version: "0.0.2", app: "phoenix"}), "")
+    user = HexWeb.Repo.get_by!(User, username: "eric")
```
wojtekmach: it's probably better to do it in a separate PR, but we could avoid queries by doing:

```elixir
test "release docs", %{user: user} do
```

test/hex_web/registry_builder_test.exs
```diff
@@ -30,14 +30,17 @@ defmodule HexWeb.RegistryBuilderTest do
...
+    postgrex = HexWeb.Repo.get_by!(Package, name: "postgrex")
+    decimal = HexWeb.Repo.get_by!(Package, name: "decimal")
+
+    Release.create(ex_doc, rel_meta(%{version: "0.0.1", app: "ex_doc"}), "") |> HexWeb.Repo.insert!
+    Release.create(decimal, rel_meta(%{version: "0.0.1", app: "decimal"}), "") |> HexWeb.Repo.insert!
+    Release.create(decimal, rel_meta(%{version: "0.0.2", app: "decimal", requirements: [%{name: "ex_doc", requirement: "0.0.1"}]}), "") |> HexWeb.Repo.insert!
+    Release.create(postgrex,
+      rel_meta(%{version: "0.0.2", app: "postgrex",
+                 requirements: [%{name: "decimal", requirement: "~> 0.0.1"}, %{name: "ex_doc", requirement: "0.0.1"}]}), "")
+    |> HexWeb.Repo.insert!
```
wojtekmach: this looks weird, any ideas how it should be styled?

test/models/release_test.exs
```diff
@@ -32,10 +33,15 @@ defmodule HexWeb.ReleaseTest do
...
 
-    assert {:ok, _} = Release.create(decimal, rel_meta(%{version: "0.0.1", app: "decimal"}), "")
-    assert {:ok, _} = Release.create(decimal, rel_meta(%{version: "0.0.2", app: "decimal"}), "")
-    assert {:ok, _} = Release.create(postgrex, rel_meta(%{version: "0.0.1", app: "postgrex", requirements: %{"decimal" => "~> 0.0.1"}}), "")
-    assert {:ok, _} = Release.create(ecto, rel_meta(%{version: "0.0.1", app: "ecto", requirements: %{"decimal" => "~> 0.0.2", "postgrex" => "== 0.0.1"}}), "")
+    Release.create(decimal, rel_meta(%{version: "0.0.1", app: "decimal"}), "") |> HexWeb.Repo.insert!
+    Release.create(decimal, rel_meta(%{version: "0.0.2", app: "decimal"}), "") |> HexWeb.Repo.insert!
+    Release.create(postgrex,
+                   rel_meta(%{requirements: [%{name: "decimal", requirement: "~> 0.0.1"}],
+                              app: "postgrex", version: "0.0.1"}), "") |> HexWeb.Repo.insert!
```
wojtekmach: same comment as above, need to find better way to style this

test/hex_web/registry_builder_test.exs
josevalim: Break rel_meta into a variable.

web/models/package_download.ex
josevalim: We need primary keys here in order to identify uniqueness when handling associations. If this primary key is not good enough, we will need to add one to the schema.

test/fixtures/sitemap.xml
josevalim: We changed it in Ecto 2.0. :)

web/controllers/api/docs_controller.ex
josevalim: You can also cheat and have a column named `:logger_id` that is a virtual field in the Audit schema. This way you can always treat it as an association.

josevalim: I would say "cheating" would be preferred because you don't actually need multi in this case. Everything would be done via the associations.

josevalim:
> revert changes to Requirement related to dependency resolution - make sure we catch dependency conflicts. Might be tricky because we add one requirement at a time with cast_assoc

Can you expand on this? How can I help?

> Clean up HexWeb.Release.requirements/1 if the beta supports %{model | ...} select syntax

We don't support this yet.

test/controllers/api/docs_controller_test.exs
ericmj: I agree, there are lots of improvements that can be done to the testing. It has a lot of legacy debt from moving between different ecto versions and moving to phoenix.

ericmj: A library for generating test data can probably help.

web/models/package_download.ex
```diff
@@ -2,11 +2,11 @@ defmodule HexWeb.PackageDownload do
...
+  @primary_key {:view, :string, []}
 
   schema "package_downloads" do
     belongs_to :package, Package, references: :id
-    field :view, :string
+    # field :view, :string
```
ericmj: I may be missing something but I think this is wrong, actually I don't know how it works :). We should add a fake primary key to the postgres view that we use as source for this schema and the use that as primary key in the schema.

ericmj:
Looks very good, let us know if we can help with anything.

wojtekmach:
Thanks! I am out for a few days so I'll resume working on it and respond to all feedback probably after the weekend.

I just remembered one thing, btw. somewhere in the diff when creating a new release and assoc reqs, I needed to preload the reqs after insert - I opened an issue for it in ecto - I'd expect not to need preload assocs as they have been just created.

web/controllers/api/release_controller.ex
```diff
@@ -118,23 +117,20 @@ defmodule HexWeb.API.ReleaseController do
...
+      Ecto.Multi.new
+      |> Ecto.Multi.insert(:release, Release.create(package, params, checksum))
+      |> Ecto.Multi.run(:log, fn %{release: release} -> audit(user, "release.publish", {package, release}) end)
+
+    case HexWeb.Repo.transaction(multi) do
+      {:ok, %{release: release}} ->
         after_release(package, release.version, user, body)
 
         release = %{release | package: package}
+        release = HexWeb.Repo.preload(release, :requirements)
```
wojtekmach: I needed to preload requirements because that association was not loaded after creation when requirements were empty. In other cases it was loaded. See: https://github.com/elixir-lang/ecto/issues/1305

web/models/package_download.ex
wojtekmach: sounds good, I'll add a proper PK

web/models/requirement.ex
```diff
@@ -4,120 +4,43 @@ defmodule HexWeb.Requirement do
...
-              optional: optional,
-              dependency_id: deps[dep])
-    |> HexWeb.Repo.insert!
-  end
-
   defp valid?(req) do
     is_binary(req) and match?({:ok, _}, Version.parse_requirement(req))
   end
-
-  defp resolve(requirements, config) do
```
wojtekmach: @josevalim 

> > revert changes to Requirement related to dependency resolution - make sure we catch dependency conflicts. Might be tricky because we add one requirement at a time with cast_assoc

> Can you expand on this? How can I help?

Sure, so here's the problem that I faced. I changed `Requirement` so that there's a `Requirement.changeset/2` function that is automatically being used by `cast_assoc(:requirements)` within `Release`. That helped a lot with cleaning up validations code and I got all tests to pass except for the one that catches requirements conflicts (there's a FIXME in this diff)

Requirement used to have this `create_all` function that accepted multiple requirements and under the hood it called this `resolve` function. Thus, I have hard time replicating this behaviour because I no longer have a list of requirements, but just a single requirement, and so I don't think I have enough data to pass in to `Hex.Resolver.resolve`

web/controllers/api/docs_controller.ex
wojtekmach: I'm kinda lost here, could you give me some more guidance how exactly this would work? Sorry for noob questions.

1. What would be the association between Release (or Package or User or ...) and AuditLog? has_one/has_many? Does it matter?

2. Would the association be, sort of, "virtual" and it would be just used for the cast_assoc (or something similar) and the audits table wouldn't directly reference any other table through a foreign key, right? I made the audit_logs schema not directly referencing anything else, and instead we serialize things into `params` JSON column. This could be changed, if desired, since in reality every audited change could be tied to one particular schema like Relase in this particular case.

3. How would usage of audits in controllers look like, as opposed to what I have right now? I think sharing some "wishful code" would be super helpful and it would allow me to work backwards from it.

Anyhow, thanks a lot for the feedback! I agree that if we can just rely on associations (and thus avoid extra DB calls) then it's probably better.

wojtekmach:
Hey @ericmj or @josevalim, sorry this PR was stale for quite a while. I'd need some guidance with 2 issues to move forward:

- Audit logs - https://github.com/hexpm/hex_web/pull/265#discussion_r56579055
- Release requirements - https://github.com/hexpm/hex_web/pull/265#discussion_r56464301

I'd appreciate any help with these!

web/models/requirement.ex
josevalim: Who owns the requirement and calls cast_assoc? Maybe that's the one that should fetch the requirements (try get_field) and pass to resolve and annotate the errors if there is something wrong?

web/controllers/api/docs_controller.ex
josevalim: Those are definitely not noob questions and we are glad to help. :) I was thinking about something like this:

    changeset
    |> put_assoc(:audit, %Audit{...})

The release would have has_one audit (or has_many, it does not really matter) and use a fake "logger_id" column in the Audit schema.

I have no guarantee this will work though but does it help you move forward?

wojtekmach: @josevalim thanks for the feedback, I really appreciate it.

I tried replacing the code above (Ecto.Multi etc) with the following code:

   ```elixir
    changeset =
      release
      |> Ecto.Changeset.change(has_docs: true)
      |> Ecto.Changeset.put_assoc(:package, Ecto.Changeset.change(release.package, docs_updated_at: Ecto.DateTime.utc))
      |> Ecto.Changeset.put_assoc(:audit_log, %AuditLog{action: "docs.publish", actor: user, params: %{
        package: %{name: package.name, id: package.id},
        release: %{}, # ???
      }})
    HexWeb.Repo.update!(changeset)
   ```

I finally can see the advantages of using assocs vs Ecto.Multi when the latter is not absolutely necessary - we also haven't lost readability which Multi provides.

I ran into an issue though, I got the following error:

```** (RuntimeError) attempting to cast or change association `audit_log` from `HexWeb.Release` that was not loaded```

I fixed it, by doing this:

```elixir
release = ... # fetch release from DB
release = Map.put(release, :audit_log, nil)
```

which feels kinda dirty. Any ideas of how to improve it? I suppose this line could be moved to each models `changeset` function so at least it wouldn't be spread all over the app.

As you can see I can easily grab `package.id` for this particular AuditLog, but I also want to have `release.id` which is not available until changeset is committed to the DB. You mentioned the fake `logger_id` field - could you expand on that, how exactly would it work in this particular case? Fake as in virtual field, or would it be the actual DB column? Would there be some custom `AuditLog.changeset/2` that `put_assoc` would automagically call or something like that?

josevalim: logger_id would be a virtual field. And yes, there would be a changeset function that would be invoked by the put_assoc with AuditLog.

> which feels kinda dirty. Any ideas of how to improve it? I suppose this line could be moved to each models changeset function so at least it wouldn't be spread all over the app.

Maybe we should allow setting defaults to associations. I can see many places where starting with them as nil or an empty list would be useful. Thoughts?

wojtekmach: > logger_id would be a virtual field. And yes, there would be a changeset function that would be invoked by the put_assoc with AuditLog.

I'm probably missing something, but it looks like `AuditLog.changeset` isn't invoked by doing `release_changeset.put_assoc(:audit_log)`. It would have been invoked when doing `release_changeset.cast_assoc(:audit_log)` though. I'm not sure if `cast_assoc` would work well here, as we would need to put auditing log data into params that come into the changeset.

I'm still not 100% sure how the virtual `logger_id` would have been automatically set - would Ecto automatically do that based on association informations, or is there some hook that can be used?

> Maybe we should allow setting defaults to associations. I can see many places where starting with them as nil or an empty list would be useful. Thoughts?

I think this would be helpful in this particular case, but I'm not sure about other use cases - it would be great to enumerate them first. Btw, would you allow setting just `nil` and `[]` or actual data as well, as in arbitrary records or arrays of records?

Thanks for your patience!

josevalim: Yes, you are right. put_assoc does not call changeset, sorry about that. My idea about using `logger_id` was because Ecto would complain if such field was not defined. BUT if Ecto is not complaining, than you don't need it.

> Btw, would you allow setting just nil and [] or actual data as well, as in arbitrary records or arrays of records?

I guess we should support whatever if we decide to add this feature. Would you please send a PR?

wojtekmach: > My idea about using logger_id was because Ecto would complain if such field was not defined.

Ecto didn't complain unless I tried to do e.g.: `belongs_to :logger, nil` in AuditLog schema - then it complained that the column is missing. When I added belongs_to and the virtual field, it complained that the field is defined twice. I thought I might need the `belongs_to` to get that `logger_id` automagically filled in - I need the id of the audited record in the audit log. On master, as well as on this PR, I'm manually setting the id of the audited record in `audit_log.params` after the first record is created within the same transaction.

> I guess we should support whatever if we decide to add this feature. Would you please send a PR?

Sure! I have a workaround for this PR and I can definitely put this on my TODO list to improve things in Ecto. An upstream PR would be good to discuss all edge cases you had in mind as well.

josevalim: You can pass `define_field: false` to belongs_to. This will allow you to define the field on the side as virtual. But I see that you also need the id of the audited record in the audit log, I think at the end of the day this won't work then. :( Sorry for pointing to the wrong direction.

wojtekmach: thanks for the tip about the field. And no worries, this discussion here was highly informative for me nonetheless!

web/models/requirement.ex
wojtekmach: Yeah that makes sense. I ended up with a similar structure as on master, there's again the  `Requirement.create_all` function which now works on `release_changeset` (instead of `release`) and it doesn't write anything to the DB anymore. See: 344cd54be3de28b3212cba3d0a786102442350c6

It still needs some work, as there's unnecessary 1 DB call per requirement and the logic around checking errors for conflicts got really complicated. The main issue I think is going back & forth between the way we need to display requirement errors on releases endpoint (to keep backwards compatibility with clients) as well as having some errors (e.g. invalid requirement string) on a particular package as well some to the requirements as a whole for the conflicts resolution (for conflict I ended up annotating every requirement with the same error). Anyhow, I'll keep at it, there's definitely some ways to clean it up even further!

web/models/requirement.ex
```diff
@@ -4,81 +4,79 @@ defmodule HexWeb.Requirement do
...
-  # TODO: Clean this up after with lands in ecto
+  def changeset(requirement, params \\ %{}) do
+    changeset = cast(requirement, params, ~w(name app requirement optional))
 
-  def create_all(release, requirements) do
-    requirements = normalize(requirements)
-    deps = deps(requirements)
+    name = changeset.changes.name
+    app = Map.get(changeset.changes, :app, name)
+    dep = HexWeb.Repo.get_by(Package, name: name)
```
wojtekmach: since `Requirement.changeset` is called from `Requirement.create_all` in a "loop" we'd end up with multiple queries whereas one query could do. Is there a way to optimize it? I kind of want to keep the `.changeset` function here because using `cast_assoc(:requirements)` is much easier than filling in everything manually.

wojtekmach:
@ericmj @josevalim please do another round of review whenever you get a chance. I think the part about Releases/Requirements could still be improved, so I'm looking forward to any suggestions.

The last 3 TODOs are IMHO optional and would require new Ecto beta release. The very last actually needs code in Ecto which I'm gonna get back to working on right now :)

josevalim:
Ecto RC will be out in 5 minutes. :)

wojtekmach:
I better get to it then!

ericmj:
@wojtekmach Great work! Is this ready for review?

wojtekmach:
@ericmj yes, please go ahead.

wojtekmach:
actually, I should clean up the git history before the final review and I need to rebase anyway. Will let you know when it's done.

test/fixtures/sitemap.xml
ericmj: I think this change breaks the sitemap spec so Z should be appended I think.

http://www.sitemaps.org/protocol.html#lastmoddef

web/models/release.ex
```diff
@@ -10,7 +10,7 @@ defmodule HexWeb.Release do
     timestamps
 
     belongs_to :package, Package
-    has_many :requirements, Requirement
+    has_many :requirements, Requirement, on_replace: :delete
```
ericmj: Would it be better to explicitly delete requirements instead? When is this needed?

web/models/requirement.ex
ericmj: You can query in the parent changeset function to build a lookup map you pass down here. It will look less good but we need to fix the N+1 here.

ericmj:
I added some minor comments and questions, it looks great.

web/models/release.ex
wojtekmach: I needed it for updating a release, when requirements changed (existing behaviour). See this test: https://github.com/wojtekmach/hex_web/blob/d029fc8d2f5ae11bda1188c3b1fda12e030ddc48/test/models/release_test.exs#L105.

test/fixtures/sitemap.xml
wojtekmach: Good catch, I'll investigate and fix if needed

wojtekmach: @ericmj Per https://github.com/elixir-lang/ecto/pull/1216, Ecto no longer assumes that date is stored in UTC, however we do so in HexWeb so I brought back the "Z" here: 6ab28b6906c580ac71a9a7c9bd0e7b94bef187b8

wojtekmach:
@ericmj I rebased and cleaned up commit history a little bit. Let me know if anything is missing! I'll do some more end-to-end testing.

test/models/release_test.exs
```diff
@@ -52,42 +59,50 @@ defmodule HexWeb.ReleaseTest do
...
+    assert %{requirements: [%{requirement: [{"invalid requirement: \"~> fail\"", []}]}]} =
+           Release.create(ecto, rel_meta(%{version: "0.1.1", app: "ecto", requirements: [%{name: "decimal", requirement: "~> fail"}]}), "") |> extract_errors
+           
+    assert %{requirements: [%{requirement: [{"Conflict on decimal\n  mix.exs: ~> 1.0\n", []}]}]} =
+           Release.create(ecto, rel_meta(%{version: "0.1.1", app: "ecto", requirements: [%{name: "decimal", requirement: "~> 1.0"}]}), "") |> extract_errors
+  end
 
-    # TODO: Remove double check when Hex 0.12.0 is out
-    assert "Conflict on decimal\n  mix.exs: ~> 1.0\n" == requirements or
-           "Failed to use \"decimal\" because\n  You specified ~> 1.0 in your mix.exs\n" == requirements
```
wojtekmach: I guess I accidentally removed this assertion when rebasing... could you help me figuring out why this is needed, and how to make this test fail (when the test is expecting `"Failed to use ..."`)? Does it depend on some kind of configuration or hex client version or something? Thanks!

ericmj: No worries, the next version of Hex updates the conflict resolution error messages. I will fix it later.

ericmj:
Great work @wojtekmach. Thank you! :heart:

PR #276, opened 2016-03-25
add sorting by recently updated

thousandsofthem:
IMO pretty useful


ericmj:
What's the use case?

thousandsofthem:
e.g. ability to view fresh packages, just not only version-`0.0.1`-ones

thousandsofthem:
One more thought:

how about this:
monosnap 2016-03-27 12-59-51
(where "more" is pointing to `/packages?sort=downloads` etc)
In such scenario adding sorting by `updated_at` is matter of uniformity

PR #279, opened 2016-03-26
Update travis.yml to match elixir_buildpack.config

wojtekmach:


wojtekmach:
Looks like travis doesn't yet have top 18.3, I'll reopen when it does.

PR #282, opened 2016-04-12
small div cleanup & moved stats to sidebar

mackenza:
Design suggestion:
Moved stats from main body to right sidebar. The main body should be information more "vital" to consumption of the package. The stats, though interesting, are not vital so moved them to sidebar.

I also removed the calendar icon. I liked it but it looked out of place on the sidebar. If we want to include such an icon, I would suggest putting an icon by each of "config" (something like: http://fortawesome.github.io/Font-Awesome/icon/cog/), "build tools" (something like: http://fortawesome.github.io/Font-Awesome/icon/wrench/ and "stats" to even out the graphical look. Let me know if you want me to mock this up.

I also fixed a div misplacement that made the Build Tools not at the same level as Config


ericmj:
Would you mind adding before and after screenshots for design changes? Thank you.

mackenza:
Sure... sorry it took me a while, I had to set up a Digital Ocean droplet so I could do this from work (where they force me to use Windows)...

### Original
hexweb-pkg-detail-original

### After PR
hexweb-pkg-detail-pr

pragmaticivan:
I'm fine with those changes, the stats in the main area was an suggestion for you guys I guess. cc @ericmj

wojtekmach:
the stats are not lining up when the download numbers are with different number of digits and it looks somewhat jarring. I think the previous style (with number above label) was better and consistent with the homepage, but I do like positioning it on the right side.

mackenza:
@wojtekmach good catch. I wonder if there is a better treatment than a big number on top of a label? I will play with it a bit and change the PR later tonight.

mackenza:
I have added the `</br>` so the numbers are on top of the text.

I also added the idea of the icons beside the sections. The two don't have to be done together, I just thought I would do them both at once to allow people to see the differences.

Let me know if either or both are wanted and I will update the PR.

hex-package-view-with-number-over-text-with-icons

pragmaticivan:
@mackenza Not sure about the others guys, but I think the first design looks better in this case. One of the comments about this page was to let just those elements in the sidebar because they where the main config info. If you compare those stats and dependencies have the same level of importance for example.

mackenza:
@pragmaticivan to each his own I guess. To me, sidebars are "oh yeah, here's some other info" and the main body is where the important info goes. To me, stats don't qualify as important info. Interesting yes, important to my use of the package, no. Thus, if the config info is important (and I agree it is), maybe it needs more prominence than the sidebar?

Overall it was just a suggestion on the design. If it's not in keeping with the design goals, I am cool with that and will still do my best to contribute. So feel free to do whatever makes sense with the PR.

:beers:

pragmaticivan:
@mackenza sure, that was how I came with this design, with a lot of feebacks :D that has changed a lot. And as I said, it just my opinion, it will be better with a general feedback.

ericmj:
I agree with @pragmaticivan that the stats look a bit better on the main part of the page and @pragmaticivan explained exactly our reasoning for what should and should not be in the sidebar. I do like the icons and thanks for fixing the dangling `</div>`.

Also appreciate all the input and feedback on design changes.

mackenza:
fair enough... we can agree to disagree on the stats ;)

I updated the PR to put the stats back on the main like before. I did make the calendar icon a bit smaller to match the text beside it. 

hex-package-view-stats-on-main-icons

ericmj:
@mackenza Can a smaller font be used instead of recreating the svg?

mackenza:
@ericmj not sure what you mean. Is this the calendar icon you are talking about? It's a png and I made it slightly smaller via css. I did that because the icon kinda of overwhelmed the text so making the text smaller seems not the right thing. Or am I misunderstanding you?

ericmj:
Sorry, I was confused by stat-build.svg and stat-config.svg? Is there any chance we can get them directly from the fonts instead? We use glyphicons already, can we fine similar icons from there?

pragmaticivan:
@mackenza I think we're already using https://fortawesome.github.io/Font-Awesome/.

If we add icon in this subtitle we're creating a new pattern style for subtitles. The subtitles in the center doesn't have icon and they're following the same style.

mackenza:
I am just going to close this PR. When I wrote it, the calendar icon was a png so I followed suit and made my icons images too. Now it has been replaced by font icons, which makes sense.
Also, I am not sure how introducing icons to headings is any more introducing a new pattern style than the calendar icons in the stats `<ul>`. Is this not the only `<ul>` on the page with icons?

Anyway, the only thing that stuck was the dangling <div> and I guess I can submit a PR for that or someone can just fix it up now that it's known. This PR seems useless now.

PR #287, opened 2016-04-16
Make footer sticky

Gazler:
On a larger screen on some the package pages, the footer would display
in the middle of the page. These changes make the footer behave as
before, except on a large resolution it will be stuck to the bottom of
the page.


Gazler:
hex-footer

New on the left, old on the right.

ericmj:
Thank you!

PR #318, opened 2016-04-20
Sorting info for default param, relocated tests

TakteS:
Added sorting info for nil param, relocated tests and added info for the sorting by updated_at. Also I found that sorting selectbox has not content "recently updated" option. It was planned, or forgot to add?

### DIFF
--- a/web/views/package_view.ex
+++ b/web/views/package_view.ex
@@ -1,8 +1,12 @@
 defmodule HexWeb.PackageView do
   use HexWeb.Web, :view
 
+  def show_sort_info(nil), do: "(Sorted by name)"
   def show_sort_info(param) do
-    available_params = %{"name" => "name", "inserted_at" => "recently created", "downloads" => "downloads"}
+    available_params = %{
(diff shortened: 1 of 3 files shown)


web/views/package_view.ex
```diff
@@ -1,8 +1,12 @@
 defmodule HexWeb.PackageView do
   use HexWeb.Web, :view
 
+  def show_sort_info(nil), do: "(Sorted by name)"
   def show_sort_info(param) do
-    available_params = %{"name" => "name", "inserted_at" => "recently created", "downloads" => "downloads"}
+    available_params = %{
+      "name" => "name", "inserted_at" => "recently created",
+      "downloads" => "downloads", "updated_at" => "recently updated"}
+
```
ericmj: You could do this with pattern matching like you did with the `nil` case also.

ericmj:
Awesome, thanks!

web/views/package_view.ex
TakteS: #319 And what about "recently updated" option

 in this selectbox?

PR #342, opened 2016-05-17
Fix JSON Timestamps

blatyo:
@ericmj @josevalim I've started creating the custom Ecto type for #340. To be honest, it seems like quite a lot of code just to avoid a dependency. Especially with the tests I've yet to write. Also, Ecto.DateTime has more methods I haven't implemented.

Should I continue in this direction?


blatyo:
Closing in favor of overriding the Poison encoder.

PR #348, opened 2016-05-22
Send email to multiple addresses at once

wojtekmach:
@ericmj I got SES working for myself, and I finally found the bug that I originally introduced when I made sending email to multiple addresses: `To` header was incorrectly formatted with multiple emails. If it would still make sense to send emails to multiple recipients (e.g. to all package owners) this way, I think this patch should fix it. An advantage is to make less calls to SES and avoid parallelism (I left HexWeb.Parallel code for future use, currently it's unused). Otherwise please close :-)

### DIFF
--- a/lib/hex_web/email/ses.ex
+++ b/lib/hex_web/email/ses.ex
@@ -3,8 +3,16 @@ defmodule HexWeb.Email.SES do
   require Logger
 
   def send(to, subject, body) do
-    HexWeb.Throttle.wait(HexWeb.SESThrottle)
-    do_send(to, subject, body)
+    ses_rate = Application.get_env(:hex_web, :ses_rate) |> String.to_integer
+
+    Enum.each(recipient_chunks(to, ses_rate), fn recipients ->
+      HexWeb.Throttle.wait(HexWeb.SESThrottle, length(recipients))
+      do_send(recipients, subject, body)
+    end)
+  end
+
+  defp recipient_chunks(recipients, limit) do
+    Enum.chunk(recipients, limit, limit, [])
   end
 
   defp do_send(to, subject, body) do
@@ -23,7 +31,7 @@ defmodule HexWeb.Email.SES do
     headers = [
       {"Subject", subject},
       {"From", "Hex.pm <#{from}>"},
-      {"To", to},
+      {"To", Enum.join(to, ",")},
       {"Return-Path", from}
     ]
 
@@ -39,7 +47,7 @@ defmodule HexWeb.Email.SES do
     result = :gen_smtp_client.send_blocking({from, to, email}, opts)
 
     unless is_binary(result) do
-      Logger.error(["Failed to send email to \"", to, "\" with subject \"", subject, "\""])
+      Logger.error(["Failed to send email to ", inspect(to), " with subject ", inspect(subject)])
       raise "Failed to send email"
     end
     result
--- a/lib/hex_web/throttle.ex
+++ b/lib/hex_web/throttle.ex
@@ -8,9 +8,9 @@ defmodule HexWeb.Throttle do
     GenServer.start_link(__MODULE__, new_state(opts), opts)
   end
 
-  def wait(pid, timeout \\ @timeout) do
+  def wait(pid, increment, timeout \\ @timeout) when increment >= 1 do
     try do
-      GenServer.call(pid, :run, timeout)
+      GenServer.call(pid, {:run, increment}, timeout)
     catch
       :exit, :timeout ->
         GenServer.cast(pid, {:cancel, self()})
@@ -18,10 +18,13 @@ defmodule HexWeb.Throttle do
     end
   end
 
-  def handle_call(:run, from, state) do
+  def handle_call({:run, increment}, _, %{rate: rate}) when increment > rate do
+    raise "increase must not exceed rate: #{inspect rate}, got: #{inspect increment}"
+  end
+  def handle_call({:run, increment}, from, state) do
     first_during_unit? = state.running == 0
 
-    state = %{state | waiting: :queue.in(from, state.waiting)}
+    state = %{state | waiting: :queue.in({from, increment}, state.waiting)}
     {_, state} = try_run(state)
 
     if first_during_unit?,
@@ -48,16 +51,17 @@ defmodule HexWeb.Throttle do
   end
 
   defp try_run(state) do
-    if state.running < state.rate do
-      case :queue.out(state.waiting) do
-        {{:value, from}, waiting} ->
+    case :queue.out(state.waiting) do
+      {{:value, {from, increment}}, waiting} ->
+        if state.running + increment <= state.rate do
           GenServer.reply(from, :yes)
-          {true, %{state | running: state.running+1, waiting: waiting}}
-        {:empty, _} ->
+          {true, %{state | running: state.running + increment, waiting: waiting}}
+        else
+          state = %{state | waiting: :queue.in({from, increment}, state.waiting)}
           {false, state}
-      end
-    else
-      {false, state}
+        end
+      {:empty, _} ->
+        {false, state}
     end
   end
 
--- a/web/mailer.ex
+++ b/web/mailer.ex
@@ -3,9 +3,6 @@ defmodule HexWeb.Mailer do
     assigns = [layout: {HexWeb.EmailView, "layout.html"}] ++ assigns
     body    = Phoenix.View.render(HexWeb.EmailView, template, assigns)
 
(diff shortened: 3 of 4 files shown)


ericmj:
Awesome, this will make https://github.com/hexpm/hex_web/issues/337 easier also! Do you know if Amazon counts a mail with multiple recipients as multiple mails in their rate limiting?

wojtekmach:
Yeah, it looks like it does. it takes AWS a long time do refresh the quotas on https://eu-west-1.console.aws.amazon.com/ses/home?region=eu-west-1#dashboard but email sent to N recipients is counted as N emails.

ericmj:
> Yeah, it looks like it does.

Hum, it is unclear if it also counts against your send rate. If so, then we need to update the email throttling.

wojtekmach:
send rate, as in N emails per second, right? Good point, I'll look into it.

wojtekmach:
@ericmj I have a max send rate of 14 emails/second on my SES account:

zrzut ekranu 2016-05-22 o 22 47 35

I tried sending 5 emails to 20 recipients each in parallel with Task.async/Task.await and it went through:

```
"Ok 01020154da3e3e8f-3f3f3045-0056-4f8a-ab8d-2b4c1ce66745-000000\r\n"
2016-05-22 20:53:13
"Ok 01020154da3e3e7f-e4885c4e-c54e-4324-9d4f-8b270f6e0b2a-000000\r\n"
2016-05-22 20:53:13
"Ok 01020154da3e3f08-0fb9261c-dfbe-4f62-87d4-08248be61ece-000000\r\n"
2016-05-22 20:53:13
"Ok 01020154da3e400b-42c5c9d7-0a6a-4d3f-855a-d376662993cb-000000\r\n"
2016-05-22 20:53:13
"Ok 01020154da3e41e6-11edc3dc-7847-4b0b-88fe-615efa51229c-000000\r\n"
2016-05-22 20:53:14
```

I had to wait a few minutes for these messages to show up in my inbox but that could be some gmail thing, not sure.

ericmj:
It sounds like their rate limit is affected by number of recipients (from http://docs.aws.amazon.com/ses/latest/DeveloperGuide/reach-sending-limits.html):

> You can send an email to multiple recipients as long as you have at least one email left before you reach your sending rate limit. Then, you have to wait until you build up the corresponding amount of sending rate quota before you can send again. For example, if your sending rate limit is one email per second, then you will be able to send an email with 10 recipients once every 10 seconds. However, we do not recommend that you send an email to multiple recipients in one call to SendEmail.

In my experience they can be lenient in their rate limiting if you only momentary exceed their limits. As a test can you send emails at a rate slightly above your limits but with only a single recipient to check if you are being throttled?

wojtekmach:
no problem, will do some more testing tomorrow :)

wojtekmach:
@ericmj I used this script to test rate limits: https://gist.github.com/wojtekmach/6d8e4c79a2e2193c0fc71316f6130a7b, and here are the results: https://gist.github.com/wojtekmach/0bbec0ef4dbdfab6e0fcec9d60b642ff and I eventually got rate limited. Some emails got delivered in an instant, for some I had to wait for 5-10+ mins.

ericmj:
Wow, it really takes a while for the sending rate limits to kick in. But it seems like you are only sending to a single recipient at a time so we don't know if the rate limiting is per email or per recipient.

Since it seems difficult to test this reliably I think we should update the throttler to bump with the number of recipients instead of a single bump per email. The docs do seem to indicate that rate limiting is per recipient instead of per email.

wojtekmach:
> But it seems like you are only sending to a single recipient at a time so we don't know if the rate limiting is per email or per recipient.

yes, you asked me to check for a single recipient ;-) I shared some previous logs when sending to multiple recipients.

Updating the throttler to account for number of recipients makes sense, I'll make this change :+1:

wojtekmach:
@ericmj I pushed draft of throttle changes in 2067ea1, but I'm not 100% confident in the implementation and will keep testing & iterating (I got throttle error, where in principle I shouldn't).

lib/hex_web/throttle.ex
```diff
@@ -8,21 +8,21 @@ defmodule HexWeb.Throttle do
     GenServer.start_link(__MODULE__, new_state(opts), opts)
   end
 
-  def wait(pid, timeout \\ @timeout) do
+  def wait(pid, increment \\ 1, timeout \\ @timeout) do
```
ericmj: Let's make this argument required. Multiple defaults are usually bad.

lib/hex_web/throttle.ex
```diff
@@ -47,12 +47,12 @@ defmodule HexWeb.Throttle do
     {:noreply, state}
   end
 
-  defp try_run(state) do
-    if state.running < state.rate do
+  defp try_run(state, increment) do
+    if state.running + increment - 1 < state.rate do
```
ericmj: `state.running + increment  <= state.rate` is more clear.

wojtekmach: :+1:

ericmj:
I think the throttle bug is that you don't store the increment in the `waiting` queue. So everyone that was queued waiting for it's turn has an increment of `1`.

wojtekmach:
Thanks for the feedback, indeed there was a bug there. I fixed it in b9a0b77 and it looks fine in my manual tests.

lib/hex_web/throttle.ex
```diff
@@ -8,20 +8,23 @@ defmodule HexWeb.Throttle do
...
     catch
       :exit, :timeout ->
         GenServer.cast(pid, {:cancel, self()})
         exit(:timeout)
     end
   end
 
-  def handle_call(:run, from, state) do
+  def handle_call({:run, increment}, _, %{rate: rate}) when increment > rate do
+    raise "increase must not exceed rate: #{inspect rate}, got: #{inspect increment}"
```
wojtekmach: otherwise the process "deadlocks". I tried unit test this, but I couldn't get it to work, I tried `catch_exit(Throttler.wait(...))` but no dice.

web/mailer.ex
```diff
@@ -3,9 +3,6 @@ defmodule HexWeb.Mailer do
     assigns = [layout: {HexWeb.EmailView, "layout.html"}] ++ assigns
     body    = Phoenix.View.render(HexWeb.EmailView, template, assigns)
 
-    HexWeb.Utils.task(fn ->
-      fun = fn email -> HexWeb.Email.send([email], title, body) end
-      HexWeb.Parallel.run(fun, emails)
-    end)
+    HexWeb.Email.send(emails, title, body)
```
ericmj: We probably have to chunk on the throttle rate here and send multiple emails or we will crash if we send a mail to more than 14 recipients?

ericmj:
One final comment, rest looks good. Please rebase to a single commit then we are good to go.

web/mailer.ex
wojtekmach: Good point, I fixed it in c0b0f76. I figured to put that logic in Email.SES since it depends on SES settings

lib/hex_web/email/ses.ex
```diff
@@ -3,8 +3,21 @@ defmodule HexWeb.Email.SES do
...
+      HexWeb.Throttle.wait(HexWeb.SESThrottle, length(recipients))
+      do_send(recipients, subject, body)
+    end)
+  end
+
+  defp recipient_chunks(recipients, limit) do
+    if length(recipients) <= limit do
+      [recipients]
+    else
+      Enum.chunk(recipients, limit)
```
ericmj: I was bitten by this myself today. Calling it this way will drop elements from the list, the correct way to call it is: `Enum.chunk(recipients, limit, limit, [])`.

lib/hex_web/email/ses.ex
```diff
@@ -3,8 +3,21 @@ defmodule HexWeb.Email.SES do
   require Logger
 
   def send(to, subject, body) do
-    HexWeb.Throttle.wait(HexWeb.SESThrottle)
-    do_send(to, subject, body)
+    ses_rate = Application.get_env(:hex_web, :ses_rate) |> String.to_integer
+
+    recipient_chunks(to, ses_rate)
+    |> Enum.each(fn recipients ->
```
ericmj: Do not pipe into a multi-line function call. Instead do: `Enum.each(recipient_chunks(to, ses_rate), fn recipients ->`.

lib/hex_web/email/ses.ex
wojtekmach: thanks for the tip with `chunk/4`! Fixed

ericmj:
Great job as usual! :heart:

PR #407, opened 2016-09-18
Update to Ecto 2.1.0

wojtekmach:


web/utils.ex
```diff
@@ -126,8 +136,8 @@ defmodule HexWeb.Utils do
     do: for(elem <- list, do: binarify(elem))
   def binarify(%Version{} = version),
     do: to_string(version)
-  def binarify(%Ecto.DateTime{} = dt),
-    do: Ecto.DateTime.to_iso8601(dt)
+  def binarify(%NaiveDateTime{} = dt),
+    do: dt |> Map.put(:microsecond, 0) |> NaiveDateTime.to_iso8601()
```
wojtekmach: added this for backwards combat: Ecto.DateTime.to_iso8601 didn't display microseconds

josevalim: You probably want to do `{0, 0}`.

josevalim:

web/utils.ex
wojtekmach: that's right, thanks!

PR #447, opened 2016-12-02
Add build_embedded and start_permanent to mix.exs example

josevalim:


web/templates/docs/publish.html.eex
```diff
@@ -93,6 +93,8 @@ You are required to confirm your email to access your account, a confirmation em
     [app: :postgrex,
      version: "0.1.0",
      elixir: "0.13.0",
```
josevalim: Yeah, we should probably do "~> 1.0" here. But I am generally ok with keeping it relatively old. :)

josevalim: Are you coming to Warsaw by any chance?

PR #465, opened 2017-01-04
Added render test for login view

ddrmanxbxfr:
This PR adds a simple test to make sure the login view renders without any crash.

Thanks!


wojtekmach:
That particular scenario is already covered with this [controller test](https://github.com/hexpm/hex_web/blob/master/test/controllers/login_controller_test.exs#L11) so I don't believe we need it. Thank you though!

PR #480, opened 2017-02-20
Improve test coverage on opensearch controller

ddrmanxbxfr:
Hi guys,

I've found that the OpenSearch controller wasn't unit tested at all.

Here's a controller test for it!

Thanks!

### DIFF
--- a/test/hexpm/web/controllers/opensearch_controller_test.exs
+++ b/test/hexpm/web/controllers/opensearch_controller_test.exs
@@ -1,0 +1,8 @@
+defmodule Hexpm.Web.OpenSearchControllerTest do
+  use Hexpm.ConnCase, async: true
+
+  test "opensearch" do
+    conn = get build_conn(), "/hexsearch.xml"
+    assert response(conn, 200) =~ "<Url type=\"text/html\" method=\"get\" template=\"http://localhost:4001/packages?search={searchTerms}&amp;sort=downloads\" />"
+  end
+end


test/controllers/opensearch_controller_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule HexWeb.OpensearchControllerTest do
+  use HexWeb.ConnCase, async: true
+
+  test "opensearch" do
+    conn = get build_conn(), "/hexsearch.xml"
+
+    path          = Path.join([__DIR__, "..", "fixtures"])
+    expected_body = File.read!(Path.join(path, "opensearch.xml"))
+
+    assert response(conn, 200) == expected_body
```
ericmj: We don't need to test the exact contents. You can use `=~` to match against a subset.

ddrmanxbxfr: Indeed, fixed.

Thanks!

test/controllers/opensearch_controller_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule HexWeb.OpensearchControllerTest do
+  use HexWeb.ConnCase, async: true
+
+  test "opensearch" do
+    conn = get build_conn(), "/hexsearch.xml"
+
+    path          = Path.join([__DIR__, "..", "fixtures"])
+    expected_body = File.read!(Path.join(path, "opensearch.xml"))
+
+    assert response(conn, 200) =~ expected_body
```
ericmj: We should only match against a subset. There is no need to include a fixture that is an exact copy of the expected output. We can match against something like: `<Url type="text/html" method="get" template="http://localhost:4001/packages?search={searchTerms}&amp;sort=downloads" />`.

test/controllers/opensearch_controller_test.exs
```diff
@@ -0,0 +1,12 @@
+defmodule HexWeb.OpensearchControllerTest do
```
wojtekmach: We've recently changed project structure [1]; please change this to `Hexpm.Web.OpenSearchControllerTest` and update filename too.

ericmj:
Ping? :)

ddrmanxbxfr:
:D I'm giving it a go tonight. Thanks for the recall.

test/controllers/opensearch_controller_test.exs
ddrmanxbxfr: Done. What would be the new filename ?

test/controllers/opensearch_controller_test.exs
ddrmanxbxfr: Indeed, added the matcher instead !

ddrmanxbxfr:
Moved the files and updated the test with a matcher instead of fixtures.

PR #505, opened 2017-05-30
Add TwoFactor Authentication [WIP]

jcspencer:
⚠️ work in progress ⚠️ 

TODO:
- [x] Move 2FA login to secondary page
- [x] Add *sudo mode* to 2FA dashboard page
- [ ] Add unit tests

--

**Implementation Details**

By default, a each user has a twofactor column that looks like:
```json
{
  "id": "<uuid>", 
  "type": "disabled", 
  "secret": "", 
  "enabled": false,
  "backupcodes": []
}
```
 
Where:
- `type` is the type of 2FA to use; currently, it's limited to 'disabled' and 'totp' (future implementations could include 'sms', etc.)
- `secret` is a shared secret (20 bytes - base32) used for TOTP. It is encrypted using PBES2_HMAC_SHA2 / AES_CBC_HMAC_SHA2 (described here hexpm/hexpm#255) with an app secret before we store it.
- `backupcodes` is an array of base32 strings - each 10 bytes (base32) which are formatted in the structure "xxxx-xxxx-xxxx" - these are single-use tokens that the user can use to access their account if they lose their phone, etc. These are also encrypted using PBES2_HMAC_SHA2 / AES_CBC_HMAC_SHA2.

**Setup Process**
- The secret, as well as ten backup codes are generated and stored in the database.
- A QR code is generated to allow the user to quickly setup Google Authenticator (or a similar TOTP generator), along with a secret to manually type in. After this, the QR code cannot be regenerated for security purposes.
- The user is requested to enter the current TOTP token to confirm that their client is setup properly (makes sure we don't lock anybody out!)
- If the code is correct, we toggle the `enabled` field and give the user their 10 backup codes.

**Authentication Process**
*Web*
- The login page has an additional field called `OTP`. If the code is incorrect or not included, the response is returned with the message "Two factor authentication required."

*API*
- The user must append the header `x-hex-otp` with the current OTP to the request
- If the TOTP token is either wrong or not included, the response is returned with the `x-hex-otp` header `required; type` (where type is either 'totp' or 'sms'), a 401 status code and the message "two-factor authentication required".

--

Screenshots:
screen shot 2017-05-30 at 9 35 17 pm
screen shot 2017-05-30 at 9 35 34 pm


mix.lock
```diff
@@ -16,8 +17,12 @@
...
+  "httpoison": {:hex, :httpoison, "0.11.1", "d06c571274c0e77b6cc50e548db3fd7779f611fbed6681fd60a331f66c143a0b", [:mix], [{:hackney, "~> 1.7.0", [hex: :hackney, repo: "hexpm", optional: false]}], "hexpm"},
   "idna": {:hex, :idna, "4.0.0", "10aaa9f79d0b12cf0def53038547855b91144f1bfcc0ec73494f38bb7b9c4961", [:rebar3], [], "hexpm"},
+  "inflex": {:hex, :inflex, "1.8.1", "9fa9684ff1a872eab7415c0be500cc1b7782f28da6ed75423081e75f92831b1c", [:mix], [], "hexpm"},
   "jiffy": {:hex, :jiffy, "0.14.11", "919a87d491c5a6b5e3bbc27fafedc3a0761ca0b4c405394f121f582fd4e3f0e5", [:rebar3], [], "hexpm"},
+  "joken": {:hex, :joken, "1.4.1", "16b87dcbad59dfb1e75231b9d6e504d852ce11f849efb7150a41dcad29dba8e1", [:mix], [{:jose, "~> 1.8", [hex: :jose, repo: "hexpm", optional: false]}, {:plug, "~> 1.0", [hex: :plug, repo: "hexpm", optional: true]}, {:poison, "~> 1.5 or ~> 2.0 or ~> 3.0", [hex: :poison, repo: "hexpm", optional: true]}], "hexpm"},
+  "jose": {:hex, :jose, "1.8.3", "1285151e6f3b70aa8b60c27801aed6d20f7bb9e87116fd7d1e405ea450549aad", [:mix, :rebar3], [{:base64url, "~> 0.0.1", [hex: :base64url, repo: "hexpm", optional: false]}], "hexpm"},
```
ericmj: Do we need all these dependencies? It seems only `qrcode` was added in mix.exs.

priv/repo/migrations/20170514062405_add_twofactor_to_users.exs
```diff
@@ -0,0 +1,11 @@
...
+  use Ecto.Migration
+
+  def change do
+    alter table(:users) do
+      add :twofactor, :jsonb, default: fragment("json_build_object('id', uuid_generate_v4()::text, 'enabled', false, 'type', 'disabled', 'secret', '', 'backupcodes', '[]'::json)::jsonb")
+    end
+  end
+end
+
+
```
ericmj: ✂️

assets/css/template/_dashboard.scss
```diff
@@ -108,3 +120,12 @@
...
 }
+
+.dashboard .backupcode {
+  font-family: Menlo,Monaco,Consolas,"Courier New",monospace;
+  line-height: 1.8 !important;
+  display: inline-block;
+  width: 48%;
+}
+
+
```
ericmj: ✂️

config/config.exs
```diff
@@ -20,6 +20,9 @@ config :hexpm,
   cookie_sign_salt: "lYEJ7Wc8jFwNrPke",
   cookie_encr_salt: "TZDiyTeFQ819hsC3",
 
+  totp_pbkdf2_iters: 32_768,
```
ericmj: This doesn't seem to be used?

lib/hexpm/accounts/totp.ex
```diff
@@ -0,0 +1,124 @@
+defmodule Hexpm.Accounts.TOTP do
+  alias Hexpm.Accounts.BackupCode
+
+  @app_secret Application.get_env(:hexpm, :totp_encryption_secret)
```
ericmj: Avoid putting config in module attributes unless it's necessary. It's because changing config requires recompilation when in attributes.

lib/hexpm/accounts/backupcode.ex
```diff
@@ -0,0 +1,51 @@
+defmodule Hexpm.Accounts.BackupCode do
+  @app_secret Application.get_env(:hexpm, :totp_encryption_secret)
```
ericmj: Avoid putting config in module attributes unless it's necessary. It's because changing config requires recompilation when in attributes.

lib/hexpm/crypto/public_key.ex
```diff
@@ -0,0 +1,27 @@
+defmodule Hexpm.Crypto.PublicKey do
```
ericmj: This module doesn't seem to be used.

lib/hexpm/accounts/user.ex
```diff
@@ -42,11 +43,65 @@ defmodule Hexpm.Accounts.User do
...
+  def use_twofactor_backupcode(user, code) do
+    decrypted = BackupCode.decrypt(user.twofactor.backupcodes)
+    new = List.delete(decrypted, code) |> BackupCode.encrypt
+
+    params = %{
+      :twofactor => %{
+        :id => user.twofactor.id,
+        :backupcodes => new
+      }
+    }
```
ericmj: This should be in `UserTwoFactor` and instead of `Changeset.cast` it should be `Changeset.change`.

lib/hexpm/web/templates/dashboard/twofactor.html.eex
```diff
@@ -0,0 +1,36 @@
...
+
+  <div class="col-md-9 twofactor">
+    <div class="panel panel-default">
+      <div class="panel-heading">2FA</div>
+      <div class="panel-body">
+        <%= form_for @changeset, dashboard_path(Endpoint, :toggle_twofactor), [method: :post], fn f -> %>
+          <%= inputs_for f, :twofactor, fn f -> %>
+            <div class="form-group">
+                <%= label f, :enabled, "Enabled?" %>
+                <%= checkbox f, :enabled %>
```
ericmj: Too much indentation.

lib/hexpm/web/templates/dashboard/twofactor.html.eex
```diff
@@ -0,0 +1,36 @@
...
+            </div>
+
+            <div class="form-group">
+                <%= label f, :type, "Type" %>
+                <%= select f, :type, @types %>
```
ericmj: Too much indentation.

lib/hexpm/web/templates/dashboard/twofactor.html.eex
```diff
@@ -0,0 +1,36 @@
...
+            </div>
+          <% end %>
+
+          <button type="submit" class="btn btn-primary">
+            Save
+          </button>
```
ericmj: Single-line this.

lib/hexpm/accounts/user_twofactor.ex
```diff
@@ -0,0 +1,46 @@
+defmodule Hexpm.Accounts.UserTwoFactor do
+  use Hexpm.Web, :schema
+
+  @types ~w(totp disabled)
+
+  embedded_schema do
+    field :enabled, :boolean, default: false
+    field :type, :string, default: "disabled"
```
ericmj: What is `type` here and what is the difference between it and `enabled`?

lib/hexpm/accounts/user_twofactor.ex
```diff
@@ -0,0 +1,46 @@
...
+    field :secret, :string, default: ""
```
ericmj: I don't think this default makes sense?

lib/hexpm/accounts/user_twofactor.ex
```diff
@@ -0,0 +1,46 @@
...
+    field :backupcodes, {:array, :string}, default: []
```
ericmj: Actually, we should review if any of these defaults should be used.

lib/hexpm/web/controllers/dashboard_controller.ex
```diff
@@ -133,6 +133,99 @@ defmodule Hexpm.Web.DashboardController do
...
+    changeset = User.setup_twofactor(user, %{})
+    enabled? = user.twofactor.enabled
+    backupcodes = User.backupcodes(user)
+
+    render_twofactor(conn, changeset, backupcodes, enabled?)
+  end
+
+  def toggle_twofactor(conn, params) do
+    user = conn.assigns.logged_in
+    enabled? = user.twofactor.enabled
```
ericmj: Shouldn't this be determined by the enabled parameter?

lib/hexpm/web/templates/dashboard/password.html.eex
```diff
@@ -34,3 +34,6 @@
     </div>
   </div>
 </div>
+
+<style>
+</style>
```
ericmj: ✂️

mix.exs
```diff
@@ -53,6 +53,7 @@ defmodule Hexpm.Mixfile do
      {:plug, "~> 1.2"},
      {:porcelain, "~> 2.0"},
      {:postgrex, ">= 0.0.0"},
+     {:qrcode, git: "https://gitlab.com/Pacodastre/qrcode"},
```
ericmj: Need to add `runtime: false` here on latest Elixir because qrcode is incorrectly saying this application has an application start callback.

lib/hexpm/accounts/user.ex
jcspencer: Agreed 👍

mix.lock
jcspencer: Fixed - I was testing `ex_twilio` locally. Whoops!

lib/hexpm/accounts/user_twofactor.ex
jcspencer: It's supposed to identify whether the 2FA is via TOTP or something else (SMS). Not sure whether it's worth keeping in at this point.

lib/hexpm/accounts/user_twofactor.ex
jcspencer: Probably not, considering they are set in the migration

config/config.exs
```diff
@@ -82,6 +84,9 @@ config :phoenix, :format_encoders,
   erlang: Hexpm.Web.ErlangFormat,
   json: Hexpm.Web.Jiffy
 
+config :phoenix, :filter_parameters,
```
jcspencer: Used to stop TOTP secrets and passwords from entering in the logs

jcspencer:
Thanks for the feedback @ericmj!; Still a lot of work to do.

I've dramatically simplified the crypto code - should be more suitable for Hex.pm.

The TOTP struct isn't completely necessary, but it simplifies TOTP configuration IMO. For example, if we introduce SMS-TOTP we can easily create a `%TOTP{}` struct with a longer interval.  Your call, but I felt it would simplify TOTP operations.

-

Quick question: If we're separating the login and two-factor entry, at what point do we enable the session? How should we ensure that the user has successfully password-authenticated when we get to twofactor auth?

ericmj:
> Quick question: If we're separating the login and two-factor entry, at what point do we enable the session? How should we ensure that the user has successfully password-authenticated when we get to twofactor auth?

We can store this information in the session.

jcspencer:
So do we create a session when they first login, and then enable the session once a twofactor token has also been entered?

ericmj:
> So do we create a session when they first login, and then enable the session once a twofactor token has also been entered?

Yes, we verify the login information on the first request and we store in the session that there is a pending login. When we have a successful 2F auth we change the pending login to an actual login.

jcspencer:
Still very much a WIP, but this is what the above commits achieve:

- redesigned login page
- sudo mode for 2fa dashboard actions
- split login page from 2fa prompt (through the use of a half-open session)

Here's a gif that shows all these new additions fairly concisely:

out

ericmj:
@jcspencer This is looking really good, thanks for your work! 😻

jcspencer:
The twofactor column has now been moved into it's own table, and is only created when 2FA is enabled. It is deleted when 2FA is disabled. The dashboard has also been improved, as it now reports the current 2FA state and lets the user toggle 2FA as opposed to giving them a confusing form.

I still have to write tests, but this is now a fairly complete implementation of 2FA. Please have a look through the PR and let me know if there's anything that you think needs to be addressed.

If you get a chance to test this, also note that if you are frequently disabling/reenabling 2FA, use this snippet to generate new TOTP codes rather than scanning QR codes endlessly into Google Authenticator: 

`User.get(name, [:twofactor]) |> User.totp(true) |> TOTP.code() `

ericmj:
Closing this due to lack of activity. Please open a new PR when it's closer to being ready. Thank you!

PR #551, opened 2017-08-19
Show tarball checksum on package page

gfvcastro:
Closes https://github.com/hexpm/hexpm/issues/549

### DIFF
--- a/lib/hexpm/web/templates/package/show.html.eex
+++ b/lib/hexpm/web/templates/package/show.html.eex
@@ -73,6 +73,11 @@ tools = Keyword.take(tools, compatible_tools)
       <span class="license"><%= Enum.join(licenses, ", ") %></span>
     <% end %>
 
+    <%= if @current_release.checksum do %>
+      <h3>Checksum</h3>
+      <span class="checksum"><%= @current_release.checksum %></span>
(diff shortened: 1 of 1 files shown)


lib/hexpm/web/templates/package/show.html.eex
```diff
@@ -73,6 +73,11 @@ tools = Keyword.take(tools, compatible_tools)
       <span class="license"><%= Enum.join(licenses, ", ") %></span>
     <% end %>
 
+    <%= if @current_release.checksum do %>
```
wojtekmach: we don't actually need an `if` because all releases have checksums:

```elixir
# prod
iex> from(r in Release, select: count(r.id), where: is_nil(r.checksum)) |> Repo.one
0
```

I'm gonna merge as is, and remove the `if` on master. I'm gonna also add a `NOT NULL` constraint for this field. Thank you!

wojtekmach: oops, I was looking at old DB schema. checksum is already NOT NULL.

gfvcastro: Oops, good catch! I missed that constraint.

PR #567, opened 2017-09-03
Fix typo in hex.organization docs

savonarola:
Hello!

Probably a typo :)


wojtekmach:
Thank you!

If you find more issues in task docs in the futures, they actually need to be fixed in Hex client itself. See the first line of the template: https://github.com/savonarola/hexpm/blob/master/lib/hexpm/web/templates/docs/tasks.html.eex#L1 :)

savonarola:
Oh, my bad. The Hex client is already fixed: https://github.com/hexpm/hex/commit/2307d2534b27ca0146df3dec57d07c89d2d338d2

wojtekmach:
yup, the fix is live already :)

PR #606, opened 2018-01-07
Cleans up UI of retirement message on package show page

supersimple:
This removes the alert box entirely if the reason is "other" and there is no message included.
It also removes the lead ": " when the reason is "other" and a message exists.

### DIFF
--- a/lib/hexpm/web/views/package_view.ex
+++ b/lib/hexpm/web/views/package_view.ex
@@ -114,27 +114,31 @@ defmodule Hexpm.Web.PackageView do
   end
 
   def retirement_message(retirement) do
+    reason = ReleaseRetirement.reason_text(retirement.reason)
     ["Retired package"] ++
       cond do
-        ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
-          ["; reason: ", ReleaseRetirement.reason_text(retirement.reason), " - ", retirement.message]
-        ReleaseRetirement.reason_text(retirement.reason) ->
-          ["; reason: ", ReleaseRetirement.reason_text(retirement.reason)]
+        reason && retirement.message ->
+          ["; reason: ", reason, " - ", retirement.message]
+        reason ->
+          ["; reason: ", reason]
         retirement.message ->
           ["; reason: ", retirement.message]
-        true -> []
+        true ->
+          []
       end
   end
 
   def retirement_html(retirement) do
+    reason = ReleaseRetirement.reason_text(retirement.reason)
     cond do
-      ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
-        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason), " - ", retirement.message]
-      ReleaseRetirement.reason_text(retirement.reason) ->
-        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason)]
+      reason && retirement.message ->
+        [content_tag(:strong, "Retired package;"), " reason: ", reason, " - ", retirement.message]
+      reason ->
+        [content_tag(:strong, "Retired package;"), " reason: ", reason]
       retirement.message ->
         [content_tag(:strong, "Retired package;"), " reason: ", retirement.message]
-      true -> [content_tag(:strong, "Retired package")]
+      true ->
+        [content_tag(:strong, "Retired package")]
     end
   end
 end
--- a/lib/hexpm/web/views/package_view.ex
+++ b/lib/hexpm/web/views/package_view.ex
@@ -114,12 +114,27 @@ defmodule Hexpm.Web.PackageView do
   end
 
   def retirement_message(retirement) do
-    [ReleaseRetirement.reason_text(retirement.reason)] ++
-      if(retirement.message, do: [": ", retirement.message], else: [])
+    ["Retired package"] ++
+      cond do
+        ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
+          ["; reason: ", ReleaseRetirement.reason_text(retirement.reason), " - ", retirement.message]
+        ReleaseRetirement.reason_text(retirement.reason) ->
+          ["; reason: ", ReleaseRetirement.reason_text(retirement.reason)]
+        retirement.message ->
+          ["; reason: ", retirement.message]
+        true -> []
+      end
   end
 
   def retirement_html(retirement) do
-    [content_tag(:strong, ReleaseRetirement.reason_text(retirement.reason))] ++
-      if(retirement.message, do: [": ", retirement.message], else: [])
+    cond do
+      ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
+        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason), " - ", retirement.message]
+      ReleaseRetirement.reason_text(retirement.reason) ->
+        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason)]
+      retirement.message ->
+        [content_tag(:strong, "Retired package;"), " reason: ", retirement.message]
+      true -> [content_tag(:strong, "Retired package")]
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/hexpm/web/templates/package/show.html.eex
```diff
@@ -38,9 +38,11 @@ tools = Keyword.take(tools, compatible_tools)
...
 
 <%= if retirement = @current_release && @current_release.retirement do %>
-  <div class="alert alert-danger" style="margin-top: 40px" role="alert">
-    <%= retirement_html(retirement) %>
-  </div>
+  <%= unless retirement_message_empty?(retirement) do %>
+    <div class="alert alert-danger" style="margin-top: 40px" role="alert">
+      <%= retirement_html(retirement) %>
+    </div>
+  <% end %>
```
ericmj: We should still always show an message but we need to improve its formatting in some cases of course.

supersimple: Sure. How about "This package has been retired." ?

ericmj: Maybe something like this: "Retired package, reason: #{reason} - #{message}"?

supersimple: Sure. In the case of "other" with no message, do you think it should read, `Retired Package, reason: other` or just `Retired Package` ?
It seems the formatter is making "Other" nil in many cases.

ericmj: Just "Retired package" is probably better in that case 👍

supersimple: Sounds good. I'll push those changes up later today.

lib/hexpm/web/views/package_view.ex
```diff
@@ -114,12 +114,27 @@ defmodule Hexpm.Web.PackageView do
...
-    [content_tag(:strong, ReleaseRetirement.reason_text(retirement.reason))] ++
-      if(retirement.message, do: [": ", retirement.message], else: [])
+    cond do
+      ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
+        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason), " - ", retirement.message]
+      ReleaseRetirement.reason_text(retirement.reason) ->
+        [content_tag(:strong, "Retired package;"), " reason: ", ReleaseRetirement.reason_text(retirement.reason)]
+      retirement.message ->
+        [content_tag(:strong, "Retired package;"), " reason: ", retirement.message]
+      true -> [content_tag(:strong, "Retired package")]
```
ericmj: Line break after `->`.

test/hexpm/web/views/package_view_test.exs
```diff
@@ -107,4 +107,51 @@ defmodule Hexpm.Web.PackageViewTest do
...
+      retirement = %{reason: "security", message: nil}
+      assert PackageView.retirement_message(retirement) == ["Retired package", "; reason: ", "Security issue"]
+    end
+  end
+
+  describe "retirement_html/1" do
+    test "reason is 'other', message contains text" do
+      retirement = %{reason: "other", message: "something went terribly wrong"}
+      assert PackageView.retirement_html(retirement) ==
+        [{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]}, " reason: ", "something went terribly wrong"]
```
ericmj: You can use `Phoenix.HTML.safe_to_string/1` to return a proper string. This way we don't rely on phoenix internals.

lib/hexpm/web/views/package_view.ex
```diff
@@ -114,12 +114,27 @@ defmodule Hexpm.Web.PackageView do
   end
 
   def retirement_message(retirement) do
-    [ReleaseRetirement.reason_text(retirement.reason)] ++
-      if(retirement.message, do: [": ", retirement.message], else: [])
+    ["Retired package"] ++
+      cond do
+        ReleaseRetirement.reason_text(retirement.reason) && retirement.message ->
```
ericmj: Bind `ReleaseRetirement.reason_text/1` result in a variable since we call it so many times.

test/hexpm/web/views/package_view_test.exs
supersimple: Sorry, I am not as familiar with PHX. From what I can tell in the documentation, I would apply that function to `{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]}`, which means it should be in the `retirement_html/1` function, not in the test. Is that correct?

supersimple: or, i could map the `PackageView.retirement_html(retirement)` result and apply the function to each element on the array, and then join it here. TBH - I am not sure what the template is expecting.

ericmj: The template expects `{:safe, ...}`. You should call `Phoenix.HTML.safe_to_string/1` in the test.

test/hexpm/web/views/package_view_test.exs
```diff
@@ -107,4 +107,51 @@ defmodule Hexpm.Web.PackageViewTest do
...
+      retirement = %{reason: "security", message: nil}
+      assert PackageView.retirement_message(retirement) == ["Retired package", "; reason: ", "Security issue"]
+    end
+  end
+
+  describe "retirement_html/1" do
+    test "reason is 'other', message contains text" do
+      retirement = PackageView.retirement_html(%{reason: "other", message: "something went terribly wrong"})
+      message = Enum.map_join(retirement, fn(x) -> if is_tuple(x), do: Phoenix.HTML.safe_to_string(x), else: x end)
+      assert message == "<strong>Retired package;</strong> reason: something went terribly wrong"
```
ericmj: This should also work:

```elixir
retirement = PackageView.retirement_html(%{reason: "other", message: "something went terribly wrong"})
assert Phoenix.HTML.safe_to_string(retirement) == "<strong>Retired package;</strong> reason: something went terribly wrong"
```

supersimple: in that case, retirement actually returns something like this: `[{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]}, " reason: ", "something went terribly wrong"]`

`Phoenix.HTML.safe_to_string` throws an error because it is expecting a tuple with the first element being the atom `:safe`. This code passes an array and it throws an exception. That is why i m doing the map, to apply the function to the tuple(s) and ignore the strings.

perhaps i have an older/newer version of phx installed that acts differently than what you expect.

supersimple: Here is the exact output from an IEx
```
iex(1)> retirement = [{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]}, " reason: ", "something went terribly wrong"]
[{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]},
 " reason: ", "something went terribly wrong"]
iex(2)> Phoenix.HTML.safe_to_string(retirement)
** (FunctionClauseError) no function clause matching in Phoenix.HTML.safe_to_string/1    
    
    The following arguments were given to Phoenix.HTML.safe_to_string/1:
    
        # 1
        [{:safe, [60, "strong", [], 62, "Retired package;", 60, 47, "strong", 62]},
     " reason: ", "something went terribly wrong"]
    
    Attempted function clauses (showing 1 out of 1):
    
        def safe_to_string({:safe, iodata})
    
    (phoenix_html) lib/phoenix_html.ex:160: Phoenix.HTML.safe_to_string/1
```

ericmj: Okay, I assumed `safe_to_string/1` accepted anything that was accepted in safe templates, sorry about that.

We can create a local function in the test module that does the `map_join` you have here and call that in all tests.

wojtekmach:
Another approach would be to break it apart a bit:

zrzut ekranu 2018-01-09 o 19 05 48

zrzut ekranu 2018-01-09 o 19 06 11

zrzut ekranu 2018-01-09 o 19 06 29

I'm fine with how it looks on the branch though, thanks!

supersimple:
that looks nice too. Although it does leave even more empty pink space on a wider screen.

wojtekmach:
Thank you!

PR #612, opened 2018-01-11
Add html_url to /api/users/me

### LINKED ISSUE #602: Add html_url to /api/users/me owned_packages payload
We need to return `html_url` to use it in `mix hex.owner packages` [1]

This is tricky as currently it's a map:

```
  "owned_packages" => %{
    "gardien" => "http://localhost:4001/api/repos/durga_transport/packages/gardien",
    "mtproto" => "http://localhost:4001/api/packages/mtproto"
  },
```

and so changing the shape will break clients. We should deprecate `owned_packages` in API payload and introduce `packages` which is a list of maps.

[1] https://github.com/hexpm/hex/blob/v0.17.1/lib/mix/tasks/hex.owner.ex#L123

Created from hexpm/hex#465.

wojtekmach:
Closes #602

### DIFF
--- a/lib/hexpm/web/views/api/user_view.ex
+++ b/lib/hexpm/web/views/api/user_view.ex
@@ -25,6 +25,7 @@ defmodule Hexpm.Web.API.UserView do
       updated_at: user.updated_at,
     }
     |> include_if_loaded(:owned_packages, user.owned_packages, &owned_packages/1)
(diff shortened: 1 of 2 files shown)


wojtekmach:
`/api/users/me` isn't documented at all so at least I won't have to mention deprecated stuff there, but yes, I'll add docs too.

PR #616, opened 2018-01-12
adds retirement info to packge endpoint

supersimple:
In order to finish https://github.com/hexpm/hex/issues/480 this change needs to be made tot he API response. It now adds information on retirements for each version to the json response.
screen shot 2018-01-12 at 1 18 57 pm

### DIFF
--- a/lib/hexpm/web/views/api/package_view.ex
+++ b/lib/hexpm/web/views/api/package_view.ex
@@ -37,8 +37,8 @@ defmodule Hexpm.Web.API.PackageView do
   end
   defp group_downloads(package), do: package
 
-  defp group_retirements(%{retirements: retirement} = package) do
-    Map.put(package, :retirements, Enum.reduce(retirement, %{}, &Map.merge(&1, &2)))
+  defp group_retirements(%{retirements: retirements} = package) do
+    Map.put(package, :retirements, Enum.reduce(retirements, %{}, &Map.merge(&1, &2)))
   end
   defp group_retirements(package), do: package
 end
--- a/lib/hexpm/web/views/api/package_view.ex
+++ b/lib/hexpm/web/views/api/package_view.ex
@@ -1,6 +1,6 @@
 defmodule Hexpm.Web.API.PackageView do
   use Hexpm.Web, :view
-  alias Hexpm.Web.API.{DownloadView, ReleaseView, UserView}
+  alias Hexpm.Web.API.{DownloadView, ReleaseView, RetirementView, UserView}
 
   def render("index." <> _, %{packages: packages}) do
     render_many(packages, __MODULE__, "show")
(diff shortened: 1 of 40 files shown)


lib/hexpm/web/views/api/retirement_view.ex
```diff
@@ -4,11 +4,21 @@ defmodule Hexpm.Web.API.RetirementView do
...
+  end
 
   def render("show", %{retirement: retirement}) do
     %{
       message: retirement.message,
       reason: retirement.reason,
     }
   end
+
+  def render("minimal", %{retirement: %{retirement: retirement}}) when is_nil(retirement), do: %{}
```
wojtekmach: fyi: `%{retirement: retirement}) when is_nil(retirement)` can be written simply as `%{retirement: nil})`

lib/hexpm/web/views/api/retirement_view.ex
```diff
@@ -4,11 +4,21 @@ defmodule Hexpm.Web.API.RetirementView do
...
+  def render("minimal", %{retirement: %{retirement: retirement, version: version}}) do
```
wojtekmach: I think there's a couple minor issues with this implementation:

1. it's called minimal, which in other views means less data. In this particular case it has the same data as `show.json` plus version so I think it might be a bit confusing in the future.
2. `%{retirement: %{retirement: retirement` looks odd, it should really be: `%{release: %{retirement: retirement`

how about:

```elixir
def render("minimal", %{retirement: %{retirement: retirement, version: version}}) do
```
Alternatively we could have:

```elixir
defmodule RetirementView do
  def render("releases." <> _, %{releases: releases}) do
    releases
    |> Enum.filter(& &1.retirement)
    |> Map.new(&{&1.version, render_one(&1.retirement, __MODULE__, "show")})
  end

  # OR:

  def render("package.", %{package: package}) do
    package.releases
    |> ...
  end
```

but either approach won't work out of the box with `include_if_loaded` helper function.

@ericmj other ideas how to best implement it?

supersimple: I pushed up a couple changes based on your feedback. I will wait to see what Eric says for implementation recommendations.

lib/hexpm/web/views/api/package_view.ex
```diff
@@ -25,15 +25,20 @@ defmodule Hexpm.Web.API.PackageView do
...
 
   defp group_downloads(%{downloads: downloads} = package) do
     Map.put(package, :downloads, Enum.reduce(downloads, %{}, &Map.merge(&1, &2)))
   end
-  defp group_downloads(package) do
-    package
+  defp group_downloads(package), do: package
+
+  defp group_retirements(%{retirement: retirement} = package) do
+    Map.put(package, :retirement, Enum.reduce(retirement, %{}, &Map.merge(&1, &2)))
```
ericmj: This should be the plural "retirements" since we can have multiple retirements for one package.

supersimple: screen shot 2018-01-13 at 9 55 57 am

I have pushed that change.

supersimple: now that i look at it, would it make more sense for retirements to be a list instead of a map?

supersimple: Now that I tried that, the list might not be a great solution since the keys are strings. If we did choose a list, I think the version would have to go back into the map along with the reason and message information.
screen shot 2018-01-13 at 10 18 44 am

ericmj:
Thank you!

PR #620, opened 2018-01-16
Avoid FOUC on Firefox

michallepicki:
This fixes https://github.com/hexpm/hexpm/issues/619


wojtekmach:
the fix is live, thank you!

PR #654, opened 2018-04-12
Fixes 404 page does not show you as logged in

### LINKED ISSUE #651: 404 page does not show you as logged in

pragmaticivan:
fixes #651 

Not sure if that would be the best solution, but works very well.

<a href="https://cl.ly/341e2d1y1G1F" target="_blank"></a>

### DIFF
--- a/lib/hexpm/web/views/error_view.ex
+++ b/lib/hexpm/web/views/error_view.ex
@@ -8,7 +8,8 @@ defmodule Hexpm.Web.ErrorView do
       error: true,
       status: status,
       message: message(status),
-      container: "container error-view"
+      container: "container error-view",
+      current_user: assigns.conn.assigns.current_user
     )
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
It is generally not advisable to perform actions on the error pages, as you are not sure what caused the error in the first place. For example, [GitHub's 404](https://github.com/odsajdlasjdklsakjdamg) does not show any login information as well.

ericmj:
I think @josevalim is right, because even with this change we will not show you as logged in if the 404 was raised in a Plug.Exception since the conn assigns are lost.

ericmj:
I think we should change the error page to something similar to github's. Sorry for misleading you @pragmaticivan.

pragmaticivan:
I agree, what about removing the header (navbar) ?

ericmj:
> I agree, what about removing the header (navbar) ?

We can try that or maybe keeping the header but only show the logo so you have a link back to the start page. Check what you think looks best.

pragmaticivan:
404page

what about that?

ericmj:
Looks great! Thank you.

pragmaticivan:
Closing in favor of https://github.com/hexpm/hexpm/pull/657

PR #661, opened 2018-04-24
Update rebar3_publish.html.md to be more clear about app metadata file

starbelly:
This is a small change, but I felt the existing doc was a bit confusing. It read as though you needed both `src/<yourapp>.app.src` and `.app.src` (i.e. in root of project).


wojtekmach:
Thank you!

PR #683, opened 2018-06-02
Api keys last use info

hubertpompecki:
Logs usage info for API keys as requested in #667.

The usage information is displayed in the admin dashboard and exposed in the api:
image
```
{
	"authing_key": true,
	"inserted_at": "2018-05-28T16:14:30.351815",
	"last_ip": "127.0.0.1",
	"last_used_at": "2018-06-02T11:31:25.708323",
	"last_user_agent": "insomnia/5.14.9",
	"name": "test key",
	"permissions": [
		{
			"domain": "api",
			"resource": "read"
		}
	],
	"revoked_at": null,
	"secret": null,
	"updated_at": "2018-06-02T11:31:25.734587",
	"url": "http://localhost:4000/api/keys/test%20key"
}
```


lib/hexpm/web/views/api/key_view.ex
```diff
@@ -20,6 +20,9 @@ defmodule Hexpm.Web.API.KeyView do
       authing_key: !!(authing_key && key.id == authing_key.id),
       secret: key.user_secret,
       permissions: render_many(key.permissions, KeyPermissionView, "show.json"),
+      last_used_at: key.last_used_at,
+      last_ip: key.last_ip,
+      last_user_agent: key.last_user_agent,
```
ericmj: What do you think about returning it as:

```elixir
last_use: %{
  timestamp: ...,
  ip: ...,
  user_agent: ...
}
```

lib/hexpm/web/templates/dashboard/key/index.html.eex
```diff
@@ -32,6 +32,15 @@
...
             </tr>
+            <tr>
+              <td colspan=3>
+                <small>
+                  Last used: <%= formatted_used_at(key.last_used_at) %><br>
+                  Last IP: <%= key.last_ip || "N/A" %><br>
+                  Last user agent: <%= key.last_user_agent || "N/A" %>
+                </small>
+              </td>
+            </tr>
```
ericmj: Can you show how it would look like if we added a "Last use" column and printed it as: "2018-06-01 09:40:00, 127.0.0.1, Hex/0.17.2"?

hubertpompecki: Hi, it just about fits in desktop but breaks mobile I am afraid:
image
image

ericmj: I think it looks okay on mobile, it is fine to line break. Please change the formatting to "2018-06-01, 127.0.0.1, Hex/0.17.2" and we are good to go.

hubertpompecki: Hi, I have made the change although I still think the layout does not look good on small mobiles. The menu shrinks to fit the screen but the "Keys" sections becomes too wide to fit.

iPhone 6/7/8 Plus in Chrome dev tools look fine:

image

However, on iPhone SE (dev tools) the menu fits on the screen but the Keys sections doesn't and can't be scrolled sideways:

image

The latter is also the case on a Huawei P10 Lite which used to test locally.

Let me know if you think it is good as it is or if you'd like to try a different approach.

ericmj:
@hubertpompecki I moved the last_use to an embeds on Key and moved the display to use popover. Thanks for your contribution!

PR #717, opened 2018-09-03
Simplify Forwarded plug.

nashby:
It's not like something important but I'm using hexpm project to learn Elixir/Phoenix so I decided to give this small refactoring a shot. Please let me know if you don't want/don't have time for in this project.

P.S https://github.com/hexpm/hexpm/compare/master...nashby:improve/forwarded-plug?expand=1#diff-7b5ce1a1b93cf4098cabff52d7e248f7R13 shouldn't it be `List.first()` here? I thought that client's ip is the first one in the list of `x-forwarded-for` ips.


wojtekmach:
Great, thank you!

PR #718, opened 2018-09-15
Fix footer issues, allow footer to be of non-fixed height

juhalehtonen:
Remove absolute positioning, forced height, and overflow:scroll from the footer to fix issues where:
- Not all content would be displayed on mobile devices
- Users would be forced to scroll footer separately on mobile
- Footer would be forced to a fixed height, limiting flow

Prior to this PR, the footer would forcibly display scrollbars at all screen sizes, which are especially visible on Windows (macOS and some others hide them system-wide). This is fixed by simply dropping the declaration and letting the content flow & the footer itself flow naturally.

This PR also fixes an issue where the forced height of the footer made mobile browsing unpleasant, as the user would be forced to scroll the footer separately from the rest of the viewport.

Lastly, we drop the absolute positioning of the footer in favour of generally letting the page flow be more natural. To compensate for the spacing that was previously applied with a margin on the body element, the footer has been given some margin-top to create spacing between it and the content area.

To ensure the footer would still remain at the end of the page regardless of content above it, we introduce a flexbox-based solution. This has been tested to work on IE11 and up, with IE10 simply not applying this positioning-related effect (thus displaying regular non-flexbox document flow).

### DIFF
--- a/assets/css/app.scss
+++ b/assets/css/app.scss
@@ -11,6 +11,7 @@
 
 @import "template/page";
 @import "template/blog";
+@import "template/content-wrapper";
 @import "template/dashboard";
 @import "template/docs";
 @import "template/home";
--- a/assets/css/base/_general.scss
+++ b/assets/css/base/_general.scss
@@ -1,12 +1,15 @@
 html {
   position: relative;
+  height: 100%;
   min-height: 100%;
 }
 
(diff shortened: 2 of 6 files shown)


wojtekmach:
Thanks you!

PR #772, opened 2018-12-31
Validate version before attempting to cast

starbelly:
This PR resolves the issue of the api returning a 500 in the case of a non-semantic version supplied for creation or update.

### DIFF
--- a/lib/hexpm/repository/releases.ex
+++ b/lib/hexpm/repository/releases.ex
@@ -159,38 +159,40 @@ defmodule Hexpm.Repository.Releases do
   defp create_release(multi, package, user, checksum, meta) do
     version = meta["version"]
 
-    params = %{
-      "app" => meta["app"],
-      "version" => version,
-      "requirements" => normalize_requirements(meta["requirements"]),
-      "meta" => meta
-    }
+    # Validate version manually to avoid an Ecto.Query.CastError exception
+    # which would return an opaque 400 HTTP status
+    case Version.parse(version) do
+      {:ok, version} ->
+        params = %{
+          "app" => meta["app"],
+          "version" => version,
+          "requirements" => normalize_requirements(meta["requirements"]),
+          "meta" => meta
+        }
 
-    case Version.parse(version) do 
-      :error ->
-        hex_ver = %{version: Hexpm.Version}
-        change = Ecto.Changeset.cast({%{}, hex_ver}, %{version: version}, ~w(version)a)
-        Ecto.Multi.error(multi, :version, change)
-      {:ok, _} -> 
         release = package && Repo.get_by(assoc(package, :releases), version: version)
 
         multi
-        |> Multi.insert_or_update(:release, fn %{
-                                             package: package,
-                                             reserved_packages: reserved_packages
-                                           } ->
-        changeset =
-          if release do
-            %{release | package: package}
-            |> preload([:requirements, :publisher])
-            |> Release.update(user, params, checksum)
-          else
-            Release.build(package, user, params, checksum)
-          end
+        |> Multi.insert_or_update(:release, fn changes ->
+          %{package: package, reserved_packages: reserved_packages} = changes
 
-        validate_reserved_version(changeset, reserved_packages)
-      end)
-      |> Multi.run(:action, fn _, _ -> {:ok, if(release, do: :update, else: :insert)} end)
+          changeset =
+            if release do
+              %{release | package: package}
+              |> preload([:requirements, :publisher])
+              |> Release.update(user, params, checksum)
+            else
+              Release.build(package, user, params, checksum)
+            end
+
+          validate_reserved_version(changeset, reserved_packages)
+        end)
+        |> Multi.run(:action, fn _, _ -> {:ok, if(release, do: :update, else: :insert)} end)
+
+      :error ->
+        params = %{version: Hexpm.Version}
+        change = Ecto.Changeset.cast({%{}, params}, %{version: version}, ~w(version)a)
+        Ecto.Multi.error(multi, :version, change)
     end
   end
 
--- a/lib/hexpm/repository/releases.ex
+++ b/lib/hexpm/repository/releases.ex
@@ -166,25 +166,32 @@ defmodule Hexpm.Repository.Releases do
       "meta" => meta
     }
 
-    release = package && Repo.get_by(assoc(package, :releases), version: version)
+    case Version.parse(version) do 
+      :error ->
+        hex_ver = %{version: Hexpm.Version}
+        change = Ecto.Changeset.cast({%{}, hex_ver}, %{version: version}, ~w(version)a)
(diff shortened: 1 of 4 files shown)


test/hexpm_web/controllers/api/release_controller_test.exs
```diff
@@ -49,6 +49,27 @@ defmodule HexpmWeb.API.ReleaseControllerTest do
...
+      conn =
+        build_conn()
+        |> put_req_header("content-type", "application/octet-stream")
+        |> put_req_header("authorization", key_for(user))
+        |> post("api/packages/#{meta.name}/releases", create_tar(meta, []))
+     
+      assert conn.status == 422
+      result = json_response(conn, 422)
+      assert result["message"] =~ "Validation error" 
+      assert result["errors"] =~ "is not semantic"
```
wojtekmach: If I keep your test changes but revert production code changes I get these 2 failures:

```
 oss/hexpm[pr-772]% mix test
................................................*.............................................................................................................................................................................................................................................................................................

  1) test POST /api/packages/:name/releases create release and new package fails with invalid version (HexpmWeb.API.ReleaseControllerTest)
     test/hexpm_web/controllers/api/release_controller_test.exs:52
     ** (FunctionClauseError) no function clause matching in Kernel.=~/2

     The following arguments were given to Kernel.=~/2:

         # 1
         %{"version" => "is invalid SemVer"}

         # 2
         "is not semantic"

     Attempted function clauses (showing 3 out of 3):

         def left =~ "" when is_binary(left)
         def left =~ right when is_binary(left) and is_binary(right)
         def left =~ right when is_binary(left)

     code: assert result["errors"] =~ "is not semantic"
     stacktrace:
       (elixir) lib/kernel.ex:1947: Kernel.=~/2
       test/hexpm_web/controllers/api/release_controller_test.exs:68: (test)

............

  2) test POST /api/packages/:name/releases update package fails when version is invalid (HexpmWeb.API.ReleaseControllerTest)
     test/hexpm_web/controllers/api/release_controller_test.exs:90
     ** (Ecto.Query.CastError) deps/ecto/lib/ecto/repo/queryable.ex:370: value `"1.0-dev"` in `where` cannot be cast to type Hexpm.Version in query:

     from r in Hexpm.Repository.Release,
       where: r.package_id == ^10243,
       where: r.version == ^"1.0-dev",
       select: r

     stacktrace:
       (elixir) lib/enum.ex:1934: Enum."-reduce/3-lists^foldl/2-0-"/3
       (elixir) lib/enum.ex:1426: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir) lib/enum.ex:1426: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (elixir) lib/enum.ex:1934: Enum."-reduce/3-lists^foldl/2-0-"/3
       (ecto) lib/ecto/repo/queryable.ex:132: Ecto.Repo.Queryable.execute/4
       (ecto) lib/ecto/repo/queryable.ex:18: Ecto.Repo.Queryable.all/3
       (ecto) lib/ecto/repo/queryable.ex:66: Ecto.Repo.Queryable.one/3
       (hexpm) lib/hexpm/repository/releases.ex:169: Hexpm.Repository.Releases.create_release/5
       (hexpm) lib/hexpm/repository/releases.ex:46: Hexpm.Repository.Releases.publish/7
       (hexpm) lib/hexpm_web/controllers/api/release_controller.ex:118: HexpmWeb.API.ReleaseController.handle_tarball/5
       (hexpm) lib/hexpm_web/controllers/api/release_controller.ex:1: HexpmWeb.API.ReleaseController.action/2

.........................................................

Finished in 4.1 seconds
405 tests, 2 failures, 1 skipped
```

See the 1st failure. `result["errors"]` should be a map and as you can see the validation already works on master when creating the release.

FYI: the way it works is we use custom `Hexpm.Version` type for the `version` field and we have code in controller helpers that transforms it's error message to `"is invalid SemVer"`.

There's still a bug on updating the release exposed by your 2nd test, good catch!

lib/hexpm/repository/releases.ex
```diff
@@ -166,25 +166,30 @@ defmodule Hexpm.Repository.Releases do
       "meta" => meta
     }
 
-    release = package && Repo.get_by(assoc(package, :releases), version: version)
-
-    multi
-    |> Multi.insert_or_update(:release, fn %{
+    case Version.parse(version) do 
+      :error ->
+        Ecto.Multi.error(multi, :version, "is not semantic")
```
wojtekmach: See comment below about tests and shape of errors.

If we return a changeset here then the rest of the system will know how to handle it. Not sure what's the nicest way to do that though, perhaps a schema less changeset like below?

```elixir
iex(1)> Ecto.Changeset.cast({%{}, %{version: Hexpm.Version}}, %{version: "1.0-dev"}, ~w(version)a)
#Ecto.Changeset<
  action: nil,
  changes: %{},
  errors: [version: {"is invalid", [type: Hexpm.Version, validation: :cast]}],
  data: %{},
  valid?: false
>
```

Thoughts?

starbelly: Yeah, I didn't feel good about wedging it in there like that and agree.

test/hexpm_web/controllers/api/release_controller_test.exs
starbelly: aye, so let's revert to "is invalid SemVer" then, missed that.

test/hexpm_web/controllers/api/release_controller_test.exs
```diff
@@ -49,6 +49,27 @@ defmodule HexpmWeb.API.ReleaseControllerTest do
       assert log.params["release"]["version"] == "1.0.0"
     end
 
+    test "create release and new package fails with invalid version", %{user: user} do
```
wojtekmach: could you remove this test? version validation is already tested on a lower level here:

https://github.com/hexpm/hexpm/blob/f9b81e779974050ec80196c8a22bb3e0d331b696/test/hexpm/repository/release_test.exs#L381

Could you replace the controller test for update with a similar lower level test? I think we can similarly extend the existing one.

starbelly: Yup.

test/hexpm/repository/release_test.exs
```diff
@@ -483,6 +483,31 @@ defmodule Hexpm.Repository.ReleaseTest do
...
+      )
+      |> Hexpm.Repo.insert!()
+
+    params =
+      params(%{
+        app: package2.name,
+        version: "1.0",
+      })
+
+      assert %Ecto.Changeset{} = error =  Release.update(%{release | package: package2}, publisher, params, "")
```
starbelly: This is causing the build to fail. Yet if I change this to match on `"is invalid Simver"` it fails locally.

PR #813, opened 2019-06-01
Check password strength against haveibeenpwned

supersimple:
Screen Shot 2019-05-31 at 4 21 27 PM

When the user logs in, updates their password, or creates a new password, the controller calls out to haveibeenpwned and checks if this password is potentially weak. If so, displays a warning flash message.

The API is not throttled at all, so that should not be an issue. I added timeouts to the hackney call so this check will not block for a long time in the controllers. if it does not respond in 500ms, we treat the password as secure and hopefully we catch them next time around.

No action is required with the warning. The user will just see it each time they log in until the problem is remedied.

I have not made any changes to the web API, so the CLI will not give this same feedback (yet.) I would like to look into that in the near future.

As always, feedback welcomed.


lib/hexpm/accounts/pwned/pwned.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Hexpm.Accounts.Pwned do
+  @moduledoc """
+    This module acts as an interface to the haveibeenpwned API
```
wojtekmach: nitpick: don't indent docs

lib/hexpm/accounts/pwned/hexpm.ex
```diff
@@ -0,0 +1,57 @@
+defmodule Hexpm.Accounts.Pwned.Hexpm do
+  @behaviour Hexpm.Accounts.Pwned
+
+  @base_url "https://api.pwnedpasswords.com/"
+  @weakness_threshold 1
+  @timeout 500
+
+  @spec password_breached?(String.t()) :: boolean
```
wojtekmach: 👍

lib/hexpm_web/controllers/controller_helpers.ex
```diff
@@ -325,6 +325,10 @@ defmodule HexpmWeb.ControllerHelpers do
   def auth_error_message(:unconfirmed),
     do: "Email has not been verified yet. You can resend the verification email below."
 
+  def password_breached_message(),
+    do: "The password you provided has previously been breached.
+      To increase your security, please change your password."
```
wojtekmach: I think this is a little bit too terse; wdyt about at least adding something like "according to https://haveibeenpwned.com"? Or instead maybe we'd document this check in docs (mention stuff like what is havebeenpwned.com, that we don't send passwords in plaintext to them etc) and link to that docs page from the flash message? I'm happy to help with copy for the docs if we're gonna move in that direction.

lib/hexpm_web/controllers/dashboard/password_controller.ex
```diff
@@ -13,8 +13,11 @@ defmodule HexpmWeb.Dashboard.PasswordController do
 
     case Users.update_password(user, params["user"], audit: audit_data(conn)) do
       {:ok, _user} ->
+        breached? = Hexpm.Accounts.Pwned.password_breached?(params["user"]["password"])
```
wojtekmach: > The API is not throttled at all, so that should not be an issue. I added timeouts to the hackney call so this check will not block for a long time in the controllers. if it does not respond in 500ms, we treat the password as secure and hopefully we catch them next time around.

I think this is a good start.

I'd like to propose a different approach, perhaps as another iteration on this feature: when we log in (or change password), we make async request to check the password so that we never block. And when that request finishes and the password is pwned, we'd mark the user as pwned and show the flash message on all pages, not just when they're logging in. And if the async request failed we can try again with a backoff etc. (We could extend that even further, if request failed we could mark the user in DB appropriately, so we could keep checking even if node is restarted.)

WDYT?

assets/package-lock.json
```diff
@@ -0,0 +1,7633 @@
+{
```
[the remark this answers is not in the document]
supersimple: Will do

config/config.exs
```diff
@@ -7,7 +7,8 @@ config :hexpm,
   support_email: "support@hex.pm",
   store_impl: Hexpm.Store.Local,
   cdn_impl: Hexpm.CDN.Local,
-  billing_impl: Hexpm.Billing.Local
+  billing_impl: Hexpm.Billing.Local,
+  pwned_impl: Hexpm.Accounts.Pwned.Hexpm
```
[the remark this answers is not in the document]
supersimple: Sure thing.

lib/hexpm/accounts/pwned/pwned.ex
```diff
@@ -0,0 +1,12 @@
+defmodule Hexpm.Accounts.Pwned do
```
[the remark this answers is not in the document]
supersimple: Sounds good

lib/hexpm_web/controllers/dashboard/password_controller.ex
supersimple: My approach is philosophically lenient. I would prefer to show the message sparingly unless we decide we really want to enforce stronger passwords (as opposed to encouraging them.) 

I considered a nonblocking call, but I think it's hard to reconcile that with this permissive approach. But the time the call responded the user could have advanced to the next page. It would also require websockets (at least the way I am imagining it.)

We should decide as a group how permissive or strict we want to be, and I think the correct approach will be clear after that.

How does that sound?

wojtekmach: Yeah, I guess it might better to either be lenient or enforce strong password, not try to nag people into changing them. Definitely looking forward to hearing from others.

lib/hexpm/accounts/pwned/hexpm.ex
```diff
@@ -0,0 +1,57 @@
...
+    |> hash_password()
+    |> occurrences_of_hash()
+    |> Kernel.>=(@weakness_threshold)
+  end
+
+  defp hash_password(string_password) do
+    :sha
+    |> :crypto.hash(string_password)
+    |> Base.encode16()
+    |> String.upcase()
```
michalmuskala: By default `Base.encode16` uses the upcase alphabet already, there's no need to upcase it afterwords again.

lib/hexpm/accounts/pwned/hexpm.ex
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  defp occurrences_of_hash(<<searchable_range::bytes-5, remainder::binary>>) do
+    searchable_range
+    |> range()
+    |> Enum.find(fn data ->
+      [hash | _t] = String.split(data, ":")
+      hash == remainder
+    end)
+    |> number_of_occurrences()
```
michalmuskala: Since the same string is split here and couple lines below, would it make sense to "pre-split" it? I think it could lead to some cleaner code:
```
searchable_range
|> range()
|> Enum.map(&String.split(&1, ":"))
|> Enum.find_value("0", fn
  [^remainder, occurrences] -> occurrences
  _ -> nil
end)
|> try_parse_or_zero()
```
Also - is it possible the API responds with something that is not an integer? If not, `try_parse_or_zero` could be replaced with a simpler `String.to_integer`.

lib/hexpm/accounts/pwned/hexpm.ex
```diff
@@ -0,0 +1,57 @@
...
+    |> String.upcase()
+  end
+
+  defp range(searchable_range) do
+    url = @base_url <> "range/#{searchable_range}"
+    headers = [{"User-Agent", "hexpm"}]
+
+    case :hackney.get(url, headers, "", connect_timeout: @timeout, recv_timeout: @timeout) do
+      {:ok, 200, _headers, ref} ->
+        {:ok, body} = :hackney.body(ref)
```
michalmuskala: If we read all of the body unconditionally anyway, I think passing the `with_body` option to hackney call might be simpler.

lib/hexpm/accounts/pwned/hexpm.ex
supersimple: the API documentation says it will always return an int after the colon.
On the other hand, there are cases when no matches are found and the response contains to colons at all. 
The reason I ended up with the [h | t] pattern match was to accommodate those cases.
(the one i saw was a string like "The range provided was invalid")

lib/hexpm/accounts/pwned/hexpm.ex
supersimple: I'll check that out, I wasnt aware of it.
It's going to take me a while to get used to looking at Erlang docs 🤪

lib/hexpm_web/controllers/controller_helpers.ex
supersimple: If we added a link to the docs/faqs in the flash message perhaps we could be a lot more expressive as to what this message is all about.
This looks similar to what GitHub does:
pwned

lib/hexpm/accounts/pwned/pwned.ex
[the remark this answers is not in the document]
supersimple: I made this change, following some of the existing naming conventions.

lib/hexpm/accounts/pwned/hexpm.ex
michalmuskala: The code I proposed should work well with those cases, since we just ignore any lists other than two elements there.

supersimple: Nice! Thanks for the suggestion.

supersimple:
Screen Shot 2019-06-01 at 6 55 05 PM

lib/hexpm_web/templates/docs/faq.html.md
```diff
@@ -9,6 +9,12 @@ username we recommend [creating a new account](/signup).
 
 You can transfer ownership of your packages using the [`hex.owner`](https://hexdocs.pm/hex/Mix.Tasks.Hex.Owner.html) task.
 
+#### Password security
```
wojtekmach: This is awesome 👍 

Can we link to this section from the flash message?

lib/hexpm/pwned/hexpm.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Hexpm.Pwned.Hexpm do
```
wojtekmach: nitpick: what do you think about calling this `Hexpm.Pwned.HaveIBeenPwned`? 

We have a similar module `Hexpm.Billing.Hexpm` but the idea there was it's using `https://billing.hex.pm` as opposed to a local fake adapter.

lib/hexpm_web/templates/docs/faq.html.md
supersimple: yeah, let me figure out how to do that.

lib/hexpm/pwned/hexpm.ex
supersimple: oh yes, good idea ;)

supersimple:
Screen Shot 2019-07-04 at 1 30 14 PM

PR #816, opened 2019-06-05
Update mix.exs file

1ternal:
Due to the fake.ex and factory.ex files were both moved to the `lib/hexpm` folder, can we just remove this line?


wojtekmach:
Good catch, thank you!

1ternal:
😄 
thank you  @wojtekmach

PR #826, opened 2019-06-25
Show audit logs in package show page

dsdshcym:
This is what it looks like for now:
image

Hi @ericmj! I'm creating this PR to get some feedback:
1. The style of this list needs to be changed. Any ideas?
2. The wording for different audit_log actions also need to be changed. Specifically, shall we provide more info for different actions? Like which owner is added for `owner.add`, etc.

### DIFF
--- a/assets/css/template/_package-view.scss
+++ b/assets/css/template/_package-view.scss
@@ -111,6 +111,16 @@
 .audit-logs-list {
   padding: 0;
   list-style: none;
+
+  li {
+    @include font-size(18);
+    margin: 0 0 10px;
+    overflow: hidden;
+    font-weight: 300;
+    line-height: 1.06;  // Just enough for e.g. the bottom of a "g" not to be cut off.
+    color: #000;
+    text-overflow: ellipsis;
+  }
 }
 
 .user-avatar {
--- a/assets/css/template/_package-view.scss
+++ b/assets/css/template/_package-view.scss
@@ -108,6 +108,11 @@
   }
 }
 
+.audit-logs-list {
+  padding: 0;
+  list-style: none;
+}
+
 .user-avatar {
   width: 30px;
   margin: 0 7px;
--- a/lib/hexpm_web/views/package_view.ex
+++ b/lib/hexpm_web/views/package_view.ex
@@ -157,35 +157,39 @@ defmodule HexpmWeb.PackageView do
     end
   end
 
+  def pretty_date(datetime) do
+    "#{DateTime.to_date(datetime)}"
+  end
+
   def humanize_audit_log_info(%{action: "docs.publish"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Publish doc"
+    "Publish doc"
   end
 
   def humanize_audit_log_info(%{action: "docs.revert"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Revert doc"
+    "Revert doc"
   end
 
   def humanize_audit_log_info(%{action: "owner.add"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Add owner"
+    "Add owner"
   end
 
   def humanize_audit_log_info(%{action: "owner.remove"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Remove owner"
+    "Remove owner"
   end
 
   def humanize_audit_log_info(%{action: "release.publish"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Publish release"
+    "Publish release"
   end
 
   def humanize_audit_log_info(%{action: "release.revert"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Revert release"
+    "Revert release"
   end
 
   def humanize_audit_log_info(%{action: "release.retire"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Retire release"
+    "Retire release"
   end
 
   def humanize_audit_log_info(%{action: "release.unretire"} = audit_log) do
-    "#{DateTime.to_date(audit_log.inserted_at)} Unretire release"
+    "Unretire release"
   end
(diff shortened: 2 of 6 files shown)


wojtekmach:
Thanks for starting this and sorry for the delay!

> The style of this list needs to be changed. Any ideas?

I think it's ok, I'd only remove the bullet symbol to save some space. WDYT?

Here's another idea how we could show them:

```
2019-07-25 09:01   Unretire release v0.1.0
2019-07-25 09:00   Retire release v0.1.0
```

by putting the date first we make it a little bit more structured which I think will help with scanning it. Also, I wonder if we should display down to hour/minute like above. Maybe not by default, but when we hover over the date we see the full timestamp?

There's another matter of the amount of audits logs, for packages with lots of activity it can easily be tens if not hundreds of items, and I don't think we should show them all. I think we should have a separate page where we list all of them (similar to e.g.: https://hex.pm/packages/ecto/versions), and on the package show page we only show latest 10 entries? Or maybe on the package show page we show no entries at all but just have a link to the page with all of them? Not sure, either option sounds fine to me.

> The wording for different audit_log actions also need to be changed. Specifically, shall we provide more info for different actions? Like which owner is added for owner.add, etc.

Agreed.

lib/hexpm_web/views/package_view.ex
```diff
@@ -156,4 +156,36 @@ defmodule HexpmWeb.PackageView do
         [content_tag(:strong, "Retired package")]
     end
   end
+
+  def humanize_audit_log_info(audit_log = %{action: "docs.publish"}) do
```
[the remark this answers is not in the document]
dsdshcym: Sure, fixed in aa1c95bb7e8798cb4f79368b23a3db7874d26bfa

dsdshcym:
Thanks @wojtekmach for your review!

>  I'd only remove the bullet symbol to save some space. WDYT?

Agreed

> Maybe not by default, but when we hover over the date we see the full timestamp?

I think so. I'll only show date by default. And add the hover for the full timestamp at last.

> we should have a separate page where we list all of them. and on the package show page we only show latest 10 entries

Yes, that's the plan listed in the original issue #754. I think it would be better to only add the latest 10 entries in this PR, and open another PR to add a separate page to show all audit logs. Just in case this PR becomes too large, and users can see the latest audit logs first. But I'm okay with doing both of them in a single PR. WDYT?

a quick list for the remaining tasks:
- [x] fix pattern matching style in function headers
- [ ] remove the bullet symbol
- [ ] show only latest 10 audit logs
- [ ] hover over the date to show the full timestamp
- [ ] a separate page to show all audit logs for a package

wojtekmach:
Agreed, lets keep this PR simple and have only recent 10 items and build upon it in future PRs.

dsdshcym:
Hi @wojtekmach!

I've made some updates, please check.

Here is how it looks now:
image

1. only 10 audit logs are shown in this page
2. the audit logs column is moved from sidebar to the main column, so that it has enough space to show all the related info
3. date is at the beginning now

some questions:
1. do we need to use the same date format for versions ("June 25, 2019") here?
2. it's wired to have date at the end for versions, but at the beginning for audit logs. Do we need to keep them the same? I suggest we move the date for versions to the beginning, because the lengths for audit log descriptions are varied.
3. do you have any suggestion for audit logs descriptions? If not, I will use my own judgement and let you review again.

wojtekmach:
> do we need to use the same date format for versions ("June 25, 2019") here?

> it's wired to have date at the end for versions, but at the beginning for audit logs. Do we need to keep them the same? 

I'm personally fine with the current design, if we want to make both lists consistent I'd rather move the date back to the end of the audit log. Perhaps it's not very helpful to have it at the beginning after all (e.g. GitHub has it at the end and displays relative dates) even though I like it this way :)

> If not, I will use my own judgement and let you review again.

Go for it!

assets/css/template/_package-view.scss
```diff
@@ -108,6 +108,21 @@
   }
 }
 
+.audit-logs-list {
+  padding: 0;
+  list-style: none;
+
+  li {
```
dsdshcym: This block is copy-pasted from L91, do we need to make them a mixin or something?

dsdshcym:
@wojtekmach The descriptions are [updated](https://github.com/hexpm/hexpm/pull/826/files#diff-8c29a912e2598a0e3e82be08b34f6c19R164-R246) and [tested](https://github.com/hexpm/hexpm/pull/826/files#diff-8f55faee5f572389dc81015912a3f6e0R183-R346), please check!

I also added handler for `owner.transfer` action

I tried to add the tooltip to show the full datetime but didn't make it. Maybe it can be added later, WDYT?

wojtekmach:
> I tried to add the tooltip to show the full datetime but didn't make it. Maybe it can be added later, WDYT?

sure thing, let's not focus on this right now. I'll review the PR later on. Thanks for your work on it!

dsdshcym:
Hi @wojtekmach!

I know you've must been busy with ElixirConf and other works recently. If you can check this PR again when you get some time, I'd be really appreciate it! 

Thank you!

wojtekmach:
Ill review it on Monday or so, sorry for the delay!

dsdshcym:
@wojtekmach No problem, thanks for the notice!

lib/hexpm_web/views/package_view.ex
```diff
@@ -156,4 +156,92 @@ defmodule HexpmWeb.PackageView do
...
+    "#{DateTime.to_date(datetime)}"
+  end
+
+  @doc """
+  This function turns an audit_log struct into a short description.
+
+  Please check Hexpm.Accounts.AuditLog.extract_params/2 to see all the
+  package related actions and their params structures.
+  """
+  def humanize_audit_log_info(%{action: "docs.publish"} = audit_log) do
```
ericmj: We might want to move this to the AuditLog schema module, because we will be showing audit logs in multiple places in the future.

We don't have to do this now, this is just a note for later.

wojtekmach:
Thank you!

PR #877, opened 2019-12-07
Call audit for "email.remove" in Users.remove_email/3

dsdshcym:
Found this issue when working on #876 
- it doesn't make sense to log "email.add" here
- and there are no other calls to log "email.remove" anywhere else


wojtekmach:
Great catch, thank you!

PR #882, opened 2019-12-31
setup 2fa login

starbelly:
Related to #852

 - Added two_factor_auth controller
 - Added two_factor_recovery controllery
 - Changed login_controller to send user down tfa path depending on tfa setting
 - Changed redirect_return in login_controller to accept a return param vs digging into the conn to support a return param coming from the session
 - Fixed up totp token creation for mobile devices which requires tokens that are 30 seconds ahead

### DIFF
--- a/lib/hexpm_web/controllers/login_controller.ex
+++ b/lib/hexpm_web/controllers/login_controller.ex
@@ -5,7 +5,7 @@ defmodule HexpmWeb.LoginController do
 
   def show(conn, _params) do
     if logged_in?(conn) do
-      redirect_return(conn, conn.assigns.current_user)
+      redirect_return(conn, conn.assigns.current_user, conn.params["return"])
     else
       render_show(conn)
     end
@@ -15,12 +15,7 @@ defmodule HexpmWeb.LoginController do
     case password_auth(username, password) do
       {:ok, user} ->
         breached? = Hexpm.Pwned.password_breached?(password)
-
-        conn
-        |> configure_session(renew: true)
-        |> put_session("user_id", user.id)
-        |> maybe_put_flash(breached?)
-        |> redirect_return(user)
+        login(conn, user, password_breached: breached?)
 
       {:error, reason} ->
         conn
@@ -36,14 +31,21 @@ defmodule HexpmWeb.LoginController do
     |> redirect(to: Routes.page_path(HexpmWeb.Endpoint, :index))
   end
 
-  defp redirect_return(%{params: %{"hexdocs" => organization}} = conn, user) do
+  def start_session(conn, user, return) do
+    conn
+    |> configure_session(renew: true)
+    |> put_session("user_id", user.id)
+    |> redirect_return(user, return)
+  end
+
+  defp redirect_return(%{params: %{"hexdocs" => organization}} = conn, user, return) do
     case generate_hexdocs_key(user, organization) do
       {:ok, key} ->
         docs_url =
           Application.get_env(:hexpm, :docs_url)
           |> String.replace("://", "://#{organization}.")
 
-        url = "#{docs_url}#{conn.params["return"]}?key=#{key.user_secret}"
+        url = "#{docs_url}#{return}?key=#{key.user_secret}"
         redirect(conn, external: url)
 
       {:error, _changeset} ->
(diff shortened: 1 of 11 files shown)


lib/hexpm_web/controllers/login_controller.ex
```diff
@@ -74,6 +77,20 @@ defmodule HexpmWeb.LoginController do
     )
   end
 
+  defp login(conn, %User{id: user_id, tfa_enabled: true}, password_breached: breached?) do
+    conn
+    |> configure_session(renew: true)
+    |> put_session("two_factor_user_id", user_id)
```
starbelly: TODO: Carry return params

lib/hexpm_web/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,40 @@
...
+        conn
+        |> delete_session("two_factor_user_id")
+        |> HexpmWeb.LoginController.start_session(user)
+      false ->
+        render_show(conn)
+    end
+  end
+
+  defp render_show(conn) do
+    render(
```
starbelly: TODO: Carry return params

lib/hexpm_web/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,40 @@
...
+      conn,
+      "show.html",
+      title: "Two Factor Authentication",
+      container: "container page page-xs login"
+    )
+  end
+
+  defp authenticate(conn, _opts) do
```
starbelly: Maybe should go in auth helpers module

lib/hexpm_web/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,40 @@
...
+  use HexpmWeb, :controller
+
+  plug :authenticate
+
+  def show(conn, _params), do: render_show(conn)
+
+  def create(conn, %{"code" => code}) do
+    uid = get_session(conn, "two_factor_user_id")
+    user = Hexpm.Accounts.Users.get_by_id(uid)
+    secret = user.tfa.secret
```
starbelly: Maybe should go in our two_factor_auth context module.

lib/hexpm_web/templates/two_factor_auth/show.html.eex
```diff
@@ -0,0 +1,14 @@
...
+  </div>
+
+  <div class="form-group">
+    <label for="code">Code</label>
+    <input id="code" type="text" class="form-control" name="code">
+  </div>
+
+<% end %>
+
+<span style="float: right;"><a href="">Enter a recovery code instead</a></span>
```
starbelly: Remains to be implemented.

lib/hexpm_web/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,40 @@
+defmodule HexpmWeb.TwoFactorAuthController do
+  use HexpmWeb, :controller
```
starbelly: Arguablly this could simply go in LoginController vs creating this new one.

lib/hexpm_web/controllers/login_controller.ex
starbelly: An alternative way to do this is create a completely separate session with a completely separate cookie, but there's some overhead there, and not sure there would be much gain.

lib/hexpm_web/controllers/two_factor_auth_controller.ex
```diff
@@ -0,0 +1,40 @@
...
+    uid = get_session(conn, "two_factor_user_id")
+    user = Hexpm.Accounts.Users.get_by_id(uid)
+    secret = user.tfa.secret
+    case Hexpm.Accounts.TwoFactorAuth.token_valid?(secret, code) do
+      true ->
+        conn
+        |> delete_session("two_factor_user_id")
+        |> HexpmWeb.LoginController.start_session(user)
+      false ->
+        render_show(conn)
```
starbelly: Needs a useful flash message to tell user the token they entered was invalid.

supersimple:
@starbelly is this ready to merge? there are some TODOs and comments. Assuming we are planning to refactor the whole feature branch at the end, is this code "ready"?

starbelly:
@supersimple Needs some tests but those can be added in your branch. A few comments left, but that can also be addressed in another PR or in your branch. 

 This is good to go! Happy new year 🎊

PR #888, opened 2020-01-17
Adds a link for each version on the package page to the diff with its…

supersimple:
… most recent predecessor


lib/hexpm/utils.ex
```diff
@@ -255,4 +255,18 @@ defmodule Hexpm.Utils do
...
+    all_versions
+    |> Enum.find_index(&(&1 == version))
+    |> case do
+      nil -> nil
+      version_index -> Enum.at(all_versions, version_index + 1)
+    end
+  end
+
+  def diff_html_url(package_name, version, previous_version) do
+    diff_url = Application.get_env(:hexpm, :diff_url)
```
wojtekmach:
```suggestion
    diff_url = Application.fetch_env!(:hexpm, :diff_url)
```

PR #891, opened 2020-01-19
Add package audit logs index page

dsdshcym:
This PR adds paginated audit logs page for packages:
image

and a link to this page in package show page:
image

### DIFF
--- a/lib/hexpm_web/controllers/package_controller.ex
+++ b/lib/hexpm_web/controllers/package_controller.ex
@@ -54,6 +54,42 @@ defmodule HexpmWeb.PackageController do
     # TODO: Show flash if private package and organization does not have active billing
 
     params = fixup_params(params)
+
+    access_package(conn, params, fn package, repositories ->
+      releases = Releases.all(package)
+
+      {release, type} =
+        if version = params["version"] do
+          {matching_release(releases, version), :release}
+        else
+          {Release.latest_version(releases, only_stable: true, unstable_fallback: true), :package}
+        end
+
+      if release do
+        package(conn, repositories, package, releases, release, type)
+      end || not_found(conn)
+    end)
+  end
+
+  def audit_logs(conn, params) do
+    access_package(conn, params, fn package, _ ->
+      page = Hexpm.Utils.safe_int(params["page"]) || 1
+      audit_logs = AuditLogs.all_by(package, 1, 100)
+      total_count = AuditLogs.count_by(package)
+
+      render(conn, "audit_logs.html",
+        title: "Recent Activities for #{package.name}",
+        container: "container package-view",
+        package: package,
+        audit_logs: audit_logs,
+        page: page,
+        per_page: 100,
+        total_count: total_count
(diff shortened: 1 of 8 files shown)


lib/hexpm_web/controllers/package_controller.ex
```diff
@@ -81,6 +81,32 @@ defmodule HexpmWeb.PackageController do
...
 
+  def audit_logs(conn, params) do
+    %{"repository" => repository, "name" => name} = params
+    organizations = Users.all_organizations(conn.assigns.current_user)
+    repositories = Map.new(organizations, &{&1.repository.name, &1.repository})
+
+    if repository = repositories[repository] do
+      package = repository && Packages.get(repository, name)
+
+      if package do
```
dsdshcym: Not sure whether or not to extract L84-L92 to a helper function like `access_package`, because `show/2` uses a lot of intermediate values (e.g. `repositories`)
if we use a helper function, then there's a lot values passing back-and-forth

lib/hexpm_web/templates/package/audit_logs.html.eex
```diff
@@ -0,0 +1,32 @@
+<h2 class="package-title">
+  <%= ViewHelpers.package_name(@package) %>
+  <span class="version">Recent Activities</span>
```
dsdshcym: Need a dedicated `class` for it
(Or a better style?)

ericmj:
@dsdshcym Unrelated to this PR but I noticed we don't handle all audit log actions in the `humanize_` functions. Can we add those or add a fallback that skips the log if we don't handle it?

lib/hexpm_web/controllers/package_controller.ex
ericmj: I think this is fine. The risk with helper functions is that we query things from the database we don't need.

lib/hexpm_web/templates/package/audit_logs.html.eex
```diff
@@ -0,0 +1,32 @@
...
+</h2>
```
ericmj: We should use the same style header we have for `templates/version/index.html.eex`.

lib/hexpm_web/controllers/package_controller.ex
dsdshcym: Not sure if I understand correctly, but I extracted the helper function in 9cfe301b8a6161d9d21a7575ef1e531cbdbc1b36 anyway

lib/hexpm_web/templates/package/audit_logs.html.eex
dsdshcym: updated in bcea77d0b86710f417b917862ad38adf6e90cbff

dsdshcym:
The header style changed after https://github.com/hexpm/hexpm/pull/891/commits/bcea77d0b86710f417b917862ad38adf6e90cbff:

image

ericmj:
Thank you!

PR #929, opened 2020-05-01
Fix anchor

reynir:
The anchor for the subheading "Example mix.exs file" seems to be
`example-mixexs-file` and not `example-mix-exs-file`.


wojtekmach:
Thank you!

Issue #922, opened 2020-04-19
Add link to login on 404 page


rodrigues:
When I open a private package page (e.g. https://hex.pm/packages/ORG/PACKAGE), and for some reason I got logged out of hex.pm, I get a 404 not found.

It would be nice if there was a link to login from that page (or show session indicator), so I could better distinguish a 404 from an access issue, and easily login and get back to that page.

Thanks!

supersimple:
Thank you for the suggestion @rodrigues .
If you would like to work on this, take a look at the error_view. It uses the simple_header, where most of the site uses header. The difference is that header expects to have @current_user set.

supersimple:
Closed by PR #930

PR #930, opened 2020-05-01
Use full header for error template (#922)

n7st:
Hi,

I had a few minutes spare and thought I'd have a go at a beginner issue (#922).

Switches simple_header.html.eex for the full header.html.eex and removes the redundant template. Passes `current_user` into the template's variables.

When viewing a page that doesn't exist:

1. If you're not logged in, the header displays the "Log in" link.
2. If you're not logged in, the header displays the user dropdown link.

Thanks!


lib/hexpm_web/views/error_view.ex
```diff
@@ -8,7 +8,8 @@ defmodule HexpmWeb.ErrorView do
       error: true,
       status: status,
       message: message(status),
-      container: "container error-view"
+      container: "container error-view",
+      current_user: assigns[:current_user]
```
supersimple, line 12: You will want to add this in the render on line 26 as well.

n7st: Thanks, makes sense :)

PR #935, opened 2020-05-09
updates all deps

supersimple:
This updates many of our dependencies.

image


lib/hexpm/ecto/version.ex
```diff
@@ -21,6 +21,9 @@ defmodule Hexpm.Version do
 
   def dump(%Version{} = version), do: {:ok, to_string(version)}
   def dump(version) when is_binary(version), do: {:ok, version}
+
+  def embed_as(_format), do: :self
+  def equal?(term1, term2), do: term1 == term2
```
wojtekmach, line 26: it's fine to keep it as is, but we can also replace `@behaviour Ecto.Type` with `use Ecto.Type` at the top of the module: https://github.com/elixir-ecto/ecto/blob/v3.4.3/lib/ecto/type.ex#L82

PR #937, opened 2020-05-14
Live dashboard redux

supersimple:
attempting to fix the liveview issue on staging by following the liveview install instructions (RTFM)

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -16,6 +16,14 @@ import "phoenix_html"
 //
 // Local files can be imported directly using relative
 // paths "./socket" or full ones "web/static/js/socket".
+import { Socket } from "phoenix"
+import LiveSocket from "phoenix_live_view"
+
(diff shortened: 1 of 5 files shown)


assets/js/app.js
```diff
@@ -50,21 +58,21 @@ export default class App {
       $(this).tab("show")
     })
 
-    $("[data-toggle='popover']").popover({container: "body", html: true, animation: false})
+    $("[data-toggle='popover']").popover({ container: "body", html: true, animation: false })
```
wojtekmach, line 61: nitpick: are you using some kind of js formatter? :)

lib/hexpm_web/endpoint.ex
```diff
@@ -3,6 +3,13 @@ defmodule HexpmWeb.Endpoint do
 
   plug HexpmWeb.Plugs.Forwarded
 
+  @session_options [
+    signing_salt: "NOcCmerj",
```
wojtekmach, line 7: should we use system env variable here?

PR #966, opened 2020-10-03
30 day graph

supersimple:
Adds 30 day download graph for releases

Screenshot from 2020-10-03 11-21-07

### DIFF
--- a/lib/hexpm_web/templates/package/show.html.eex
+++ b/lib/hexpm_web/templates/package/show.html.eex
@@ -77,7 +77,6 @@ tools = [
         <line x1="0" y1="<%= y %>" x2="800" y2="<%= y %>" stroke="#ddd"></line>
       <% end  %>
       <polyline fill="none" stroke="url(#grad1)" stroke-width="3" stroke-linecap="round" points="<%= points %>"></polyline>
-      <polyline fill="rgba(255, 255, 255, 0.75)" points="<%= points_fill %>"></polyline>
       <%= for {l, y} <- y_axis_labels do%>
         <text x="0" y="<%= y %>" fill="#aaa"><%= l %></text>
       <% end %>
--- a/lib/hexpm_web/templates/package/show.html.eex
+++ b/lib/hexpm_web/templates/package/show.html.eex
@@ -5,7 +5,7 @@ links = Enum.to_list(@package.meta.links || [])
 build_tools = @current_release && @current_release.meta.build_tools || []
 downloads = if @current_release.downloads, do: @current_release.downloads.downloads, else: 0
 {graph_labels, points, points_fill} = ViewHelpers.time_series_graph(@daily_graph)
-y_axis_labels = Enum.zip(graph_labels, [192, 152, 112, 72, 32])
+y_axis_labels = Enum.zip(graph_labels, [196, 156, 116, 76, 36])
 
 tools = [
   mix: "mix.exs",
@@ -81,6 +81,7 @@ tools = [
       <%= for {l, y} <- y_axis_labels do%>
         <text x="0" y="<%= y %>" fill="#aaa"><%= l %></text>
       <% end %>
+      <text x="0" y="12" fill="#aaa">Last 30 days</text>
     </svg>
   </div>
 <div class="col-md-11 with-divider no-padding">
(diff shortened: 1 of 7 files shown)


supersimple:
@ericmj @wojtekmach any thoughts?

ericmj:
@supersimple I want to play with this locally. Do you have a way to generate dummy download stats data?

supersimple:
@ericmj I have just been adding data via cli. no shortcuts :/

lib/hexpm_web/views/view_helpers.ex
```diff
@@ -354,6 +354,57 @@ defmodule HexpmWeb.ViewHelpers do
...
+  end
+
+  defp x_axis_points(total_points) do
+    # width / points captured
+    px_per_point = Float.round(800 / total_points, 2)
+    Enum.map(0..total_points, &Kernel.*(&1, px_per_point))
+  end
+
+  defp to_polyline_points(list) do
+    Enum.reduce(list, "", fn {y, x}, acc -> acc <> "#{x}, #{y} " end)
```
ericmj, line 387: Can you use `Enum.join` here?

ericmj (commented):
How does it look without the "shadow" under the line? I tried to remove it but I can't find what is creating it :).

For `packages/acme/` without a version I think we should the stats for the whole package instead of the latest release.

Overall I think this is great! 💜

lib/hexpm_web/views/view_helpers.ex
```diff
@@ -354,6 +354,57 @@ defmodule HexpmWeb.ViewHelpers do
...
+  defp y_axis_labels(min, max) do
+    div = (max - min) / 5
+
+    [
+      min,
+      round(div),
+      round(div * 2),
+      round(div * 3),
+      round(div * 4)
+    ]
```
ericmj, line 406: Should we humanize these values? Popular packages may have big numbers.

lib/hexpm_web/templates/package/show.html.eex
```diff
@@ -59,55 +61,74 @@ tools = [
...
+          <stop offset="66%" style="stop-color:#b5179e;" />
+          <stop offset="100%" style="stop-color:#f72585;" />
+        </linearGradient>
+      </defs>
+
+      <%= for y <- [40, 80, 120, 160, 200] do %>
+        <line x1="0" y1="<%= y %>" x2="800" y2="<%= y %>" stroke="#ddd"></line>
+      <% end  %>
+      <polyline fill="none" stroke="url(#grad1)" stroke-width="3" stroke-linecap="round" points="<%= points %>"></polyline>
+      <polyline fill="rgba(255, 255, 255, 0.75)" points="<%= points_fill %>"></polyline>
```
supersimple, line 80: this polyline is the "shadow" if you want to remove and see how it looks.
I'd also play around with the stroke-width on the line above to see how you like it thicker and thinner.

lib/hexpm_web/views/view_helpers.ex
supersimple: no, i dont think so. I need to reverse the tuple, and also need to add a comma between the x and y, which i dont think join can do.

ericmj: I was thinking mainly to introduce the comma, but since we are doing so many other things the reduce is probably better.

supersimple:
Screenshot from 2020-10-29 10-52-04

Here is the graph with normalized y-axis values, and no shadow.

I think we should have a way to see the downloads for all versions, but I am not sure where to put it.
I do not think the latest version page should show the all versions graph.
having 2 lines in this one graph (one for all versions, one for this version) is going to lead to some weird y-axises. you might see this version in the hundreds, and all versions in the tens of thousands (making the graph for this version uterly meaningless.

WDYT?

ericmj:
> I do not think the latest version page should show the all versions graph.

Why not? That's what we do for the other download numbers.

supersimple:
We always show a number that is for this version. Then we show totals. AFAIK this includes the latest, which is default.

I just think the behavior is confusing if the default view is all versions and all other views are for that version only.

If anything we should show both, but that means changing the y axis.

I am not holding this opinion strongly. But I am trying to think of the use case for it and what would help users.

supersimple:
@ericmj ok, so I see the benefit of showing all versions on the default page. The one thing I think is weird is that there is no real navigation to that view on the page. So, if you look at a release, the only way to get to the default page is either through the address bar directly or to go back to search.

Can you think of any solutions to that?

ericmj:
💜

PR #971, opened 2020-10-22
Correct delete_many behaviour spec

kpanic:
While skimming through the code, I just noticed this small mismatch


wojtekmach:
Thanks!

PR #1005, opened 2021-02-17
Fix github action name

thiamsantos:
The CI seems to be breaking https://github.com/hexpm/hexpm/runs/1853646204.

Updating to use the new action seems to solve the issue.


wojtekmach:
Thanks!

PR #1008, opened 2021-03-03
Unify lists of versions

wojtekmach:
Previously we used a slightly different template for the list on the
package page and on the all versions page, now we re-use the former.

Before:

image

After:

image


ericmj:
I like that we get all the links to docs/diff/preview and reuse code but I think I prefer the former look because it looks a bit strange when the list and the header are not aligned. Previously the list took the full width of the page so it didn't matter.

Can you try centering the list or changing the header?

wojtekmach:
Good call, I should have either used the previous style or have removed it. This is how it looks now, I think we can ship that:

image

ericmj (approved):
Looks great!

PR #1031, opened 2021-05-20
Drop www subdomain for rebar3 links

ferd:
Something's kind of not working great with the redirect to drop the
subdomain (which I have no control over) so this instead prevents
display issues people get when coming off of the hex website.

Reference: https://github.com/tsloughter/rebar3.org/issues/76#issuecomment-845543555


wojtekmach:
Thanks!

PR #1038, opened 2021-06-27
Ability to set public email as private

### LINKED ISSUE #1034: Add "Set as non-public" button to emails
Currently we have this:

image

to make existing email non-public, we need to add another one, make it primary, delete the old one, and re-add it.

Let's add a "Set as non-public" button.

christopherlai:
Enables the ability to set a public email as private via the dashboard (Closes #1034).

### DIFF
--- a/lib/hexpm/accounts/audit_log.ex
+++ b/lib/hexpm/accounts/audit_log.ex
@@ -152,9 +152,6 @@ defmodule Hexpm.Accounts.AuditLog do
   defp extract_params("email.public", {old_email, new_email}),
     do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
 
-  defp extract_params("email.private", {old_email, new_email}),
-    do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
-
   defp extract_params("email.gravatar", {organization, {old_email, new_email}}),
     do: %{
       organization: serialize(organization),
--- a/lib/hexpm/accounts/audit_log.ex
+++ b/lib/hexpm/accounts/audit_log.ex
@@ -149,9 +149,10 @@ defmodule Hexpm.Accounts.AuditLog do
       new_email: serialize(new_email)
     }
 
-  defp extract_params("email.private", email), do: serialize(email)
-
   defp extract_params("email.public", {old_email, new_email}),
+    do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
+
+  defp extract_params("email.private", {old_email, new_email}),
     do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
 
   defp extract_params("email.gravatar", {organization, {old_email, new_email}}),
--- a/lib/hexpm/accounts/audit_log.ex
+++ b/lib/hexpm/accounts/audit_log.ex
@@ -149,6 +149,8 @@ defmodule Hexpm.Accounts.AuditLog do
       new_email: serialize(new_email)
     }
 
+  defp extract_params("email.private", email), do: serialize(email)
+
   defp extract_params("email.public", {old_email, new_email}),
     do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
 
--- a/lib/hexpm/accounts/users.ex
+++ b/lib/hexpm/accounts/users.ex
@@ -449,24 +449,6 @@ defmodule Hexpm.Accounts.Users do
     end
   end
 
-  def private_email(user, params, audit: audit_data) do
-    public_email = Enum.find(user.emails, & &1.public)
-
-    multi =
-      if public_email.email == params["email"] do
-        Multi.new()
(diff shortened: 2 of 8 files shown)


test/hexpm_web/views/dashboard/audit_log_view_test.exs
```diff
@@ -136,6 +136,13 @@ defmodule HexpmWeb.Dashboard.AuditLogViewTest do
                "Set email test@example.com as public email"
     end
 
+    test "email.pvaite" do
```
wojtekmach, line 139:
```suggestion
    test "email.private" do
```

lib/hexpm/accounts/users.ex
```diff
@@ -448,6 +448,24 @@ defmodule Hexpm.Accounts.Users do
     end
   end
 
+  def private_email(user, params, audit: audit_data) do
+    public_email = Enum.find(user.emails, &Map.get(&1, :public))
```
wojtekmach, line 452:
```suggestion
    public_email = Enum.find(user.emails, & &1.public)
```

wojtekmach:
@christopherlai thank you for the PR!

I didn't realise we actually have this functionality already :-(

image

image

but I still think we should move forward with the button as I think that's nicer UI. WDYT @ericmj?

If so we should reuse and/or slightly change existing code. Currently it is handled here: https://github.com/hexpm/hexpm/blob/main/lib/hexpm/accounts/users.ex#L394:L396

wojtekmach:
Oh, there's another problem with the current public email toggling (your PR fixes that), both enabling and disabling looks the same in the audit log:

image

christopherlai:
> Oh, there's another problem with the current public email toggling (your PR fixes that), both enabling and disabling looks the same in the audit log:
> 
> image

I noticed that bug as well when I was working on this.

@wojtekmach I've updated the PR to use the existing functions + fixed the bug above as well.

lib/hexpm/accounts/audit_log.ex
```diff
@@ -152,6 +152,9 @@ defmodule Hexpm.Accounts.AuditLog do
   defp extract_params("email.public", {old_email, new_email}),
     do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
 
+  defp extract_params("email.private", {old_email, new_email}),
+    do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
```
wojtekmach, line 156:
```suggestion
  defp extract_params("email.private", {old_email, nil}),
    do: %{old_email: serialize(old_email), new_email: nil}
```

wojtekmach (approved):

lib/hexpm/accounts/audit_log.ex
wojtekmach: In fact we should probably store it as:

```elixir
%{email: serialize(old_email)}
```

but I think we should leave it as is to be uniform with `email.public` log entry.

lib/hexpm/accounts/users.ex
```diff
@@ -402,10 +402,11 @@ defmodule Hexpm.Accounts.Users do
   defp unset_email_flag_multi(multi, user, flag, audit: audit_data) do
     if old_email = Enum.find(user.emails, &Map.get(&1, flag)) do
       old_email_op = String.to_atom("old_#{flag}")
+      action = if flag == :public, do: :private, else: flag
 
       multi
       |> Multi.update(old_email_op, Email.toggle_flag(old_email, flag, false))
-      |> audit(audit_data, "email.#{flag}", {old_email, nil})
+      |> audit(audit_data, "email.#{action}", {old_email, nil})
```
ericmj, line 409: I think we should keep the audit log the same and instead change how we display it in the UI, so that if `new_email: nil` we say that the public email was removed/hidden.

ericmj (commented):
If we add this we should remove the option from the selector dropdown.

wojtekmach:
> *If* we add this we should remove the option from the selector dropdown.

(emphasis mine) just to be clear, you're in favour of moving forward with this? or would rather leave things as is?

ericmj:
> (emphasis mine) just to be clear, you're in favour of moving forward with this? or would rather leave things as is?

I am in favor of moving forward with this PR.

christopherlai:
@wojtekmach @ericmj what do we want to do about the profile page? Should the dropdown be removed in favor of static text? There is already a link there directing users to the Emails page to add and remove, should the email page be the canonical page to make any changes to one's emails?

lib/hexpm/accounts/users.ex
christopherlai: Addressed in https://github.com/hexpm/hexpm/pull/1038/commits/32e7621ce944e7985aef9f6f8b65b6885a7476d7

christopherlai:
@wojtekmach @ericmj I think this is good now.

ericmj (approved):

wojtekmach:
Thank you!

PR #1048, opened 2021-08-03
Do not start runtime_tools and os_mon in tests

wojtekmach:
Before this patch we had:

    ~/src/hexpm[main]% mix test
    Finished in 5.4 seconds (4.8s async, 0.6s sync)
    669 tests, 0 failures, 3 skipped

    Randomized with seed 594000
    [os_mon] memory supervisor port (memsup): Erlang has closed
    [os_mon] cpu supervisor port (cpu_sup): Erlang has closed

Now we have:

    ~/src/hexpm[main]% mix test
    Finished in 5.4 seconds (4.8s async, 0.6s sync)
    669 tests, 0 failures, 3 skipped

    Randomized with seed 594000

So just removing unnecessary logs in the test output.

It doesn't hurt to keep `runtime_tools` but I think it is not useful either.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -20,9 +20,12 @@ defmodule Hexpm.MixProject do
   def application() do
     [
       mod: {Hexpm.Application, []},
-      extra_applications: [:logger, :runtime_tools, :os_mon]
+      extra_applications: extra_applications(Mix.env())
     ]
   end
+
+  defp extra_applications(:test), do: [:logger]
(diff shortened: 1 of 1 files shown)


ericmj (approved):
Are they useful in dev or should we only start them in prod?

wojtekmach:
I think it is useful to have os_mon even in dev, for live dashboard. not sure about runtime_tools. I guess it has dbg but I never use it :D

PR #1058, opened 2021-10-07
WebAuth Proof of Concept

### LINKED ISSUE #1024: Web auth to authenticate CLI from browser
José [mentioned](https://github.com/elixir-lang/elixir/issues/10489#issuecomment-833499522) that you (the hexpm team) are interested in creating a means of authenticating the CLI from the browser. Could I get some more info please ?

Benjamin-Philip:
The following is a Proof of Concept for web auth.

Closes #1024


Benjamin-Philip:
@wojtekmach here's the pr.

ericmj:
> I have got an incomplete implementation in Schema. Whenever I run Repo.update in submit, I get the following error. I figured it might be due to the way I am migrating this, but I am unable to troubleshoot. Any ideas?

It's hard to say without seeing what's being updated and when I try to run your branch it fails with a compilation error, but it looks like postgres is expecting an id/integer where you are passing a Key schema or Key params.

I do notice that you are creating a Key with Keys.create and then embedding it in WebAuth which means it is stored in two places which we should try to avoid. Why does WebAuth need to store a Key?

Benjamin-Philip:
> If the user can retrieve the key multiple times the point of this is moot.

So, in the GenServer implementation, the request is deleted after few minutes, after which the user can't access the key.

Anyways, we'll generate the key when the user attempts to access it like you said.

Benjamin-Philip:
@ericmj, I have pushed a complete implementation of WebAuth. Please review when you get the time. :)

Benjamin-Philip:
@ericmj, also I don't plan to work on the controller so as to avoid doing redundant work in case the Schema implementation is not to your satisfaction.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,155 @@
...
+
+    case WebAuth.changeset(%WebAuth{}, request) |> Repo.insert() do
+      {:ok, _req} ->
+        %{
+          device_code: device_code,
+          user_code: user_code,
+          verification_uri: @verification_uri,
+          access_key_uri: @access_key_uri,
+          verification_expires_in: @verification_expires_in,
+          key_access_expires_in: @key_access_expires_in
```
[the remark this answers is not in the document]
Benjamin-Philip: @ericmj I disagree.

`*_expires_in` can probably be removed.

 However, `*_uri` should be retained as those will tell the user where to access and verify the key.  This will prevent hard coding in the cli and will make any client easier to maintain.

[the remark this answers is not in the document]
ericmj: The clients already needs to hard code other URLs so I don't think this adds much value.

ericmj:
@Benjamin-Philip I noticed you pushed more commits. Let us know when it's ready for second review.

Benjamin-Philip:
Sure. 

I only fixed some minor stuff and still have to split the Web auth module into a context. I also have the remove the extra information.

The thing is this and next month, I may not have much time to work on this PR.

Benjamin-Philip:
> Sure. 
> 
> I only fixed some minor stuff and still have to split the Web auth module into a context. I also have the remove the extra information.
> 
> The thing is this and next month, I may not have much time to work on this PR. So updates are going to be slow.

@ericmj, I hope there isn't a hurry?

wojtekmach:
There is no rush :)

Benjamin-Philip:
@ericmj and @wojtekmach, I have split `WebAuth` into a context and a schema. 

I'm not very sure if what I have done is the right way to go about it. (All the functions in the Context are impure, so I think it is.)

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,155 @@
+defmodule Hexpm.Accounts.WebAuth do
```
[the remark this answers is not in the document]
Benjamin-Philip: Done. Please have a look.

Benjamin-Philip:
@ericmj, and @wojtekmach, please review the `WebAuthRequest` schema and the `WebAuth` context so that I can make improvements (or plan accordingly to work on them).

Note: Changes to HexpmWeb is not yet review ready.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+  use Hexpm.Context
+
+  @moduledoc false
+
+  # A pool for storing and validating web auth requests.
+
+  alias Hexpm.Accounts.WebAuthRequest, as: Req
```
wojtekmach, line 8: nitpick: let's call it by its full name? (not least because using Req name will trip me every single time because https://hex.pm/packages/req :)) 

```suggestion
  alias Hexpm.Accounts.WebAuthRequest
```

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+
+  @doc """
+  Adds a web auth request to the pool and returns the response.
+
+  ## Params
+
+    - `scope` - Scope of the key to be generated. One of read and write.
+  """
+
+  def get_code(scope) do
```
wojtekmach, line 29:
```suggestion
  def get_code(params) do
```

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+    device_code = "foo"
+    user_code = "bar"
+
+    request = %{
+      device_code: device_code,
+      user_code: user_code,
+      scope: scope,
+      verified: false
```
wojtekmach, line 37: we should not be passing internal fields like "verified" through changeset cast. cast is only for user input. Instead, make `verified: false` the default, and later on, do:

```elixir
changeset |> put_change(:verified, true)
```

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+  ## Params
+
+  - `device_code` - The device code assigned to the client
+  """
+  def access_key(device_code) do
+    request = Req |> Repo.get_by(device_code: device_code)
+
+    case request do
+      r when r.verified == true ->
+        user = Repo.get(Hexpm.Accounts.User, request.user_id)
```
wojtekmach, line 87:
```suggestion
        user = Repo.get!(Hexpm.Accounts.User, request.user_id)
```

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,56 @@
...
+    case WebAuth.get_code(scope) do
+      {:error, _changeset} ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> json(%{"error" => "invalid scope"})
+
+      response ->
+        conn
+        |> put_status(:ok)
+        |> json(response)
```
wojtekmach, line 22: 200 is the default:

```suggestion
        json(conn, response)
```

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,56 @@
...
+
+  def submit(conn, %{"user_code" => user_code}) do
+    audit = audit_data(conn)
+    user = conn.assigns.current_user
+
+    case WebAuth.submit(user, user_code, audit) do
+      {:ok, _changeset} ->
+        conn
+        |> put_status(:ok)
+        |> redirect(to: Routes.web_auth_path(conn, :success))
```
wojtekmach, line 38: just double-checking, is it intentional to redirect with HTTP 200 as opposed to HTTP 302?

priv/repo/migrations/20211014073049_add_web_auth_requests_table.exs
```diff
@@ -0,0 +1,14 @@
+defmodule Hexpm.RepoBase.Migrations.AddWebAuthRequestsTable do
+  use Ecto.Migration
+
+  def change do
+    create table(:requests) do
+      add(:device_code, :string)
+      add(:user_code, :string)
+      add(:scope, :string)
+      add(:verified, :boolean)
+      add(:user_id, :integer)
```
wojtekmach, line 10:
```suggestion
      add(:user_id, references(:users), null: false)
```

I believe we should have `null: false` on all fields?

test/hexpm/accounts/web_auth_test.exs
```diff
@@ -0,0 +1,101 @@
...
+        response = WebAuth.get_code(scope)
+
+        assert response.device_code
+        assert response.user_code
+      end
+    end
+
+    test "returns an error on invalid scope" do
+      {status, _changeset} = WebAuth.get_code("foo")
+      assert status == :error
```
wojtekmach, line 20:
```suggestion
      assert {:error, changeset} = WebAuth.get_code("foo")
      assert errors_on(changeset)[...] == ...
```

test/hexpm/accounts/web_auth_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    end
+  end
+
+  describe "submit/3" do
+    setup [:get_code, :login]
+
+    test "returns ok on valid params", c do
+      audit_data = audit_data(c.user)
+
+      {status, _changeset} = WebAuth.submit(c.user, c.request.user_code, audit_data)
```
wojtekmach, line 30:
```suggestion
      assert {:ok, ...} = WebAuth.submit(c.user, c.request.user_code, audit_data)
```

please assert what `...` is. If it is unused in production code too, let's just return `:ok` from this function.

test/hexpm/accounts/web_auth_test.exs
```diff
@@ -0,0 +1,101 @@
...
+    setup [:get_code, :login]
+
+    test "returns a key on valid device code", c do
+      submit_code(c)
+
+      key =
+        c.request.device_code
+        |> WebAuth.access_key()
+
+      assert key.__struct__ == Hexpm.Accounts.Key
```
wojtekmach, line 51:
```suggestion
      assert %Hexpm.Accounts.Key{} = key
```

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+  """
+  def submit(user, user_code, audit) do
+    request = Req |> Repo.get_by(user_code: user_code)
+
+    if request do
+      {_user, audit_con} = audit
+
+      change = %{verified: true, user_id: user.id, audit: audit_con}
+
+      Req.changeset(request, change) |> Repo.update()
```
wojtekmach, line 69: if the DB call fails, our function returns `{:error,%Changeset{} | "invalid user code"}`. Can we do something more uniform, ie we always return a changeset or always a string error?

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,117 @@
...
+        name = "Web Auth #{device_code} key"
+
+        # Add name
+        key_params = %{@key_params | name: name}
+        # Add permissions
+        key_params = %{key_params | permissions: [%{@key_permission | resource: scope}]}
+
+        case Keys.create(user, key_params, audit: audit) do
+          {:ok, %{key: key}} ->
+            Repo.delete(request)
```
wojtekmach, line 102: we should do Keys.create and Repo.delete in the same transaction. I think using Ecto multi here would make sense.

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -0,0 +1,26 @@
...
+    field :user_code, :string
+    field :scope, :string
+    field :verified, :boolean
+    field :user_id, :integer
+    field :audit, :string
+  end
+
+  def changeset(request, params \\ %{}) do
+    request
+    |> cast(params, [:device_code, :user_code, :scope, :verified, :user_id, :audit])
```
wojtekmach, line 21: we shouldn't allow `:user_id` in the cast either. Again, cast is for user input and we don't want to accidentally ever allow someone injecting values for FKs and other internal fields.

lib/hexpm/accounts/web_auth.ex
Benjamin-Philip: Sure.

lib/hexpm_web/controllers/web_auth_controller.ex
Benjamin-Philip: No it isn't.

lib/hexpm/accounts/web_auth.ex
Benjamin-Philip: Fixed at https://github.com/hexpm/hexpm/pull/1058/commits/a83a2bf34a0389fd8b79d91b63ad093a62f97643. Have a look.

priv/repo/migrations/20211014073049_add_web_auth_requests_table.exs
Benjamin-Philip: > I believe we should have null: false on all fields?

Yes.

lib/hexpm/accounts/web_auth.ex
Benjamin-Philip: @wojtekmach, which do you prefer? Changeset or string?

test/hexpm/accounts/web_auth_test.exs
Benjamin-Philip: `...` is the response from `Repo.update` in the case of valid params.

We might have a `{:error, "invalid user code"}` in the case of invalid params.

lib/hexpm/accounts/web_auth.ex
wojtekmach: whatever is simpler to implement, probably a string.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,113 @@
...
+  """
+
+  def get_code(params) do
+    device_code = "foo"
+    user_code = "bar"
+
+    params = %{
+      device_code: device_code,
+      user_code: user_code,
+      scope: params
```
wojtekmach, line 36: sorry, I didn’t mean to simply rename scope to params. What I had in mind was params be a map, scope be one of its keys.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,113 @@
...
+  Adds a web auth request to the pool and returns the response.
+
+  ## Params
+
+    - `scope` - Scope of the key to be generated. One of read and write.
+  """
+
+  def get_code(scope) do
+    device_code = "foo"
+    user_code = "bar"
```
Benjamin-Philip, line 31: We really ought to find a way of generating these values.

Any suggestions @wojtekmach and @ericmj ?

wojtekmach: user code is easier, it should be relatively easy to type so I'd go with what GitHub has, I believe 8 letters and/or digits? Device code must have higher entropy though. I'd also check how roughly GitHub does this, how long is the secret. How about this for device code?

```elixir
iex(19)> :crypto.strong_rand_bytes(16) |> Base.url_encode64(padding: false)
"5bbMz-sRFl_udEZTUC9s3g"
```

Benjamin-Philip: > I believe 8 letters and/or digits?

Yes.

> I'd also check how roughly GitHub does this, how long is the secret.

They don't explain how they generate the code, but they do mention that it is 40 chars long.

Looking at it, I think it is SHA, but I'm not very knowledgeable in cryptography.

See [the docs](https://docs.github.com/en/developers/apps/building-oauth-apps/authorizing-oauth-apps#step-1-app-requests-the-device-and-user-verification-codes-from-github)

Benjamin-Philip: Looking at the [rfc](https://datatracker.ietf.org/doc/html/rfc8628#section-3.2),

The seemed to be just using 32 random bytes instead of 16:

```elixir
iex(13)> :crypto.strong_rand_bytes(32) |> Base.url_encode64(padding: false)
"W_hqp7AGf3FfQzipLXQTC98Gur03ex8RE9h4a4oUYCM"
```

Benjamin-Philip: For the user code, how about:

```elixir
:crypto.strong_rand_bytes(4) |> Base.hex_encode32(padding: false)
```

(I am using hex encoding to avoid `-` and `_`)

And for the hyphen in the middle

```elixir
:crypto.strong_rand_bytes(5) 
|> Base.hex_encode32(padding: false) 
|> String.split_at(4) 
|> then(fn {x, y} -> x <> "-" <> y end)
"QL0U-6B4M"
```

wojtekmach:
@Benjamin-Philip your snippets for user and device code generation are looking good to me 👍

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,122 @@
...
+
+  ## Params
+
+  - `device_code` - The device code assigned to the client
+  """
+  def access_key(device_code) do
+    request = WebAuthRequest |> Repo.get_by(device_code: device_code)
+
+    case request do
+      r when r.verified == true ->
```
DaniruKun, line 91: Is the full comparison expression really required? I think just `r when r.verified` should be sufficient.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,122 @@
...
+          {:ok, %{key: key}} ->
+            Repo.delete(request)
+
+            key
+
+          error ->
+            error
+        end
+
+      r when r.verified == false ->
```
DaniruKun, line 115: Same as above, though the guard might be redundant, since it will match automatically if the previous guard doesn't match as far as I understand.

lib/hexpm_web/router.ex
```diff
@@ -201,6 +204,13 @@ defmodule HexpmWeb.Router do
     get "/feeds/blog.xml", FeedsController, :blog
   end
 
+  scope "/login/web_auth", HexpmWeb do
+    pipe_through :api
+
+    post "/code", WebAuthController, :code
+    post "/access_key", WebAuthController, :access_key
```
DaniruKun, line 211: I would also probably implement a fallback controller for WebAuth, see https://hexdocs.pm/phoenix/Phoenix.Controller.html#action_fallback/1-examples

lib/hexpm/accounts/web_auth.ex
Benjamin-Philip: You're right!

lib/hexpm/accounts/web_auth.ex
Benjamin-Philip: Here, a `not` with a guard is necessary as we need to account for a nil value.

lib/hexpm_web/router.ex
Benjamin-Philip: Ah, I wasn't aware of fallback controllers. Thanks!

lib/hexpm/accounts/web_auth_request.ex
Benjamin-Philip: @wojtekmach,

Would you re-write the schema using a Many-to-one association?

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -0,0 +1,57 @@
...
+  @key_permission %{domain: "api", resource: nil}
+  @key_params %{name: nil, permissions: [@key_permission]}
+
+  schema "requests" do
+    field :device_code, :string
+    field :user_code, :string
+    field :scope, :string
+    field :verified, :boolean, default: false
+    belongs_to :user, Hexpm.Accounts.User
+    field :audit, :string
```
Benjamin-Philip, line 20: Is this okay for audit data? For testing we use a string, but I am not so sure if you do so in production?

Benjamin-Philip:
@wojtekmach, I have worked on all your reviews pertaining to the context/schema. Can you re-review please?

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -0,0 +1,57 @@
...
+  end
+
+  def access_key(request) do
+    user = request.user
+
+    audit = {user, request.audit}
+
+    scope = request.scope
+    device_code = request.device_code
+    name = "Web Auth #{device_code} key"
```
Benjamin-Philip, line 46: @ericmj, @wojtekmach, Is this key name okay?

This things is that the device code is 43 characters long, so:

1. It is a very long name
2. It does not indicate what device the key belongs to.
3. It may not be very easy to distinguish it from other Web Auth keys

Benjamin-Philip:
@wojtekmach and @ericmj, could review the context please?

test/hexpm_web/controllers/web_auth_controller.exs
```diff
@@ -0,0 +1,85 @@
+defmodule HexpmWeb.WebAuthControllerTest do
```
ericmj, line 1: This file needs to be named `web_auth_controller_test.exs`, otherwise tests won't run.

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,54 @@
...
+
+  # step one of device flow
+  def code(conn, %{"scope" => scope}) do
+    case WebAuth.get_code(scope) do
+      {:error, _changeset} ->
+        conn
+        |> put_status(:unprocessable_entity)
+        |> json(%{"error" => "invalid scope"})
+
+      response ->
```
ericmj, line 19:
```suggestion
      {:ok, response} ->
```

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,54 @@
...
+        json(conn, response)
```
ericmj, line 20: This should be rendered in a view `WebAuthView` and explicit select which fields to render so that we do not accidentally include private data if `WebAuth.get_code/1` returns more information in the future.

Same for the `success` and `access_token` actions.

lib/hexpm_web/router.ex
```diff
@@ -62,6 +62,9 @@ defmodule HexpmWeb.Router do
     post "/login", LoginController, :create
     post "/logout", LoginController, :delete
 
+    get "/login/web_auth", WebAuthController, :show
+    post "/login/web_auth/submit", WebAuthController, :submit
```
ericmj, line 66: I would rather move this to its own controller so we don't have a single controller handling both browser and API requests.

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,54 @@
...
+      {:error, _msg} ->
+        conn
+        |> put_status(400)
+
+      :ok ->
+        redirect(conn, to: Routes.web_auth_path(conn, :sucess))
+    end
+  end
+
+  def sucess(conn, _) do
```
ericmj, line 47: This doesn't seem to be used for anything?

priv/repo/migrations/20211014073049_add_web_auth_requests_table.exs
```diff
@@ -0,0 +1,14 @@
...
+
+  def change do
+    create table(:requests) do
+      add(:device_code, :string, null: false)
+      add(:user_code, :string, null: false)
+      add(:scope, :string, null: false)
+      add(:verified, :boolean, null: false)
+      add(:user_id, references(:users))
+      add(:audit, :string)
+    end
```
ericmj, line 12: We need indexes on the columns we query on.

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -0,0 +1,57 @@
...
+
+    # Add name
+    key_params = %{@key_params | name: name}
+    # Add permissions
+    key_params = %{key_params | permissions: [%{@key_permission | resource: scope}]}
```
ericmj, line 51:
```suggestion
    key_params = %{@key_params | name: name}
    key_params = %{key_params | permissions: [%{@key_permission | resource: scope}]}
```

ericmj (commented):
When the CLI authenticates it creates two keys, one key with read only permissions and key with write permissions so that the write key can be password protected. With web auth we don't want the user to go through the authentication flow twice so we should be able to generate multiple keys.

This is starting to look really good, thanks for your work on this. I think the next step is to start working on the CLI client to verify that we have the right approach and API for this. What do you think @Benjamin-Philip?

lib/hexpm/accounts/web_auth_request.ex
Benjamin-Philip: > I don't think authenticating through the web API requires an audit log though, we usually only log actions that change something in the system.

@ericmj, I need the Audit information to generate a key.

> Audit information is stored separately in the AuditLog schema.

I believe for tests that this is stored as a tuple in the form `{%Hexpm.Accounts.User, "test"}`.
Should I submit/work on a PR to fix this in the tests?

Benjamin-Philip:
@ericmj,

> When the CLI authenticates it creates two keys, one key with read only permissions and key with write permissions so that the write key can be password protected. With web auth we don't want the user to go through the authentication flow twice so we should be able to generate multiple keys.

I was not aware of this. Will fix.

>  This is starting to look really good, thanks for your work on this.

Thanks!

> I think the next step is to start working on the CLI client to verify that we have the right approach and API for this. 

I don't think there is anything wrong with this approach/API.

I will start working on an implementation for the mix client, and once this and the mix client PR have been merged, I will work on the rebar3 client.

Also: The Phoenix side of things is incomplete. I might have a completed version ready soon.

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -41,15 +39,16 @@ defmodule Hexpm.Accounts.WebAuthRequest do
...
-    device_code = request.device_code
-    name = "Web Auth #{device_code} key"
+    key_name = request.key_name
 
-    key_params = %{@key_params | name: name}
-    key_params = %{key_params | permissions: [%{@key_permission | resource: scope}]}
+    key_params = %{@key_params | name: key_name}
+
+    write_key = %{key_params | permissions: [%{@key_permission | resource: "write"}]}
+    read_key = %{key_params | permissions: [%{@key_permission | resource: "read"}]}
```
ericmj, line 47: What do you think about letting the user specifying what keys and with what permissions to generate? That way it's more flexible for the future.

Benjamin-Philip: > What do you think about letting the user specifying what keys and with what permissions to generate? That way it's more flexible for the future.

I think we can always work on that later. Right now, I think we need to get a minimum, viable version ready, which I think this current implementation would be once I figure out the submission/verification form. This would allow me to get a implementation ready for the client without wasting much time, rather than working on more features on the server side.

Benjamin-Philip: However, I do like the idea of specifying permissions.

Benjamin-Philip:
@ericmj, and @wojtekmach,

I am having some difficulty on displaying an error on invalid user code, since the error is a string and not a changeset.
I have committed my attempt, but I think its' not the right way of going about. Could you help?

lib/hexpm/accounts/web_auth_request.ex
ericmj: > I need the Audit information to generate a key.

Is there an issue getting the audit information when generating the key?

test/hexpm_web/controllers/web_auth_controller_test.exs
```diff
@@ -0,0 +1,133 @@
...
+      post(context.conn, Routes.web_auth_path(context.conn, :code, @test))
+      |> json_response(200)
+
+    Map.merge(context, %{request: request})
+  end
+
+  def submit(c) do
+    request = %{"user_code" => c.request["user_code"]}
+
+    post(c.conn, Routes.web_auth_path(c.conn, :submit, request))
```
ericmj, line 129: There are tests for this above so let's call the context functions instead of going through the controller.

test/hexpm_web/controllers/web_auth_controller_test.exs
```diff
@@ -0,0 +1,133 @@
...
+  def build_conn(context) do
+    Map.merge(context, %{conn: build_conn()})
+  end
+
+  def login(context) do
+    user = insert(:user)
+    organization = insert(:organization)
+    insert(:organization_user, organization: organization, user: user)
+
+    # test login
```
ericmj, line 112:
```suggestion
```

test/hexpm/accounts/web_auth_test.exs
```diff
@@ -0,0 +1,102 @@
...
+        "bad code"
+        |> WebAuth.access_key()
+
+      assert response == {:error, "invalid device code"}
+    end
+
+    test "deletes request after user has accessed", c do
+      submit_code(c)
+
+      c.request.device_code |> WebAuth.access_key()
```
ericmj, line 74: Avoid single pipes since they do not add much to the readability.

lib/hexpm/accounts/web_auth_request.ex
```diff
@@ -0,0 +1,54 @@
...
+    key_name = request.key_name
+
+    key_params = %{@key_params | name: key_name}
+
+    write_key = %{key_params | permissions: [%{@key_permission | resource: "write"}]}
+    read_key = %{key_params | permissions: [%{@key_permission | resource: "read"}]}
+
+    Multi.new()
+    |> Multi.run(:write_key_gen, fn _, _ -> Keys.create(user, write_key, audit: audit) end)
+    |> Multi.run(:read_key_gen, fn _, _ -> Keys.create(user, read_key, audit: audit) end)
```
ericmj, line 51: Schema modules should generally not call back out to context modules. Move this into the `WebAuth` context.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,110 @@
+defmodule Hexpm.Accounts.WebAuth do
+  use Hexpm.Context
+
+  @moduledoc false
+
+  # A pool for storing and validating web auth requests.
```
ericmj, line 6:
```suggestion
```

Benjamin-Philip:
@ericmj, also in the previous comments, I was referring to difficulty in making the user_code submission form.

lib/hexpm_web/router.ex
ericmj: I don't this was done.

lib/hexpm_web/controllers/web_auth_controller.ex
```diff
@@ -0,0 +1,93 @@
...
+
+  def submit(conn, %{"user_code" => user_code}) do
+    audit = audit_data(conn)
+    user = conn.assigns.current_user
+
+    case WebAuth.submit(user, user_code, audit) do
+      {:ok, _request} ->
+        conn
+        |> put_status(:ok)
+        |> json(%{"ok" => "ok"})
```
ericmj, line 30: It looks like we are returning JSON for a website request?

Benjamin-Philip: I am going to redirect to a success page on successful verification. The JSON is temporary.

ericmj:
> also in the previous comment, I was referring to difficulty in making the user_code submission form.

I think the way you're doing is great. Is there any issue with it?

Benjamin-Philip:
> > also in the previous comment, I was referring to difficulty in making the user_code submission form.
> 
> I think the way you're doing is great. Is there any issue with it?

There weren't any built-in function or a function in `ViewHelpers` that enabled me to return an error from a string - all function dealt with changesets. So, I wrote my own. I was not sure if writing a function myself was the right idea, in case there was already a function that does what I want to do.

The problem is, that my error looks different from the error passed in say the login page, as a result of me writing it by myself. I think this can be solved by tweaking the css.

So, I think there isn't any major issue with it.

Benjamin-Philip:
@ericmj, @wojtekmach,

I think I have got a working MVP/POC ready.

Things I think we need to improve on:

- Styling for the verification and success pages
- Support for organisation keys
- Anything else you think needs working on

Here's how the verification and success pages look:
Web capture_9-12-2021_144555_localhost
Web capture_9-12-2021_144456_localhost

The verification page is probably okay if left alone, but I feel that the success page looks a bit bare.

When it comes to supporting organisation keys, I think this would be possible by supporting user supplied. key creation parameters. I will work on that after we have got a basic version in production.

Benjamin-Philip:
Here is a video showing the flow:

https://user-images.githubusercontent.com/67545861/145373327-6a8e4bbf-787e-418e-aa97-781814a027b6.mp4

Note: I think the frame rate is a bit slow because I set my `ffmpeg` flags incorrectly when converting to mp4.

Benjamin-Philip:
Any suggestions/comments @ericmj?

ericmj:
> Styling for the verification and success pages

I think they are fine as-is. We are doing a full redesign of the website so don't focus too much on it. We just need to make sure the right CSS classes are applied to the validation error so that it looks consistent with the rest of the website.

> Support for organisation keys

That's not needed since users do not log in as organizations.

I will do another review later today or during the weekend.

lib/hexpm_web/controllers/web_auth_browser_controller.ex
```diff
@@ -0,0 +1,57 @@
+defmodule HexpmWeb.WebAuthBrowserController do
```
ericmj, line 1:
```suggestion
defmodule HexpmWeb.WebAuthController do
```

lib/hexpm_web/templates/web_auth_browser/show.html.eex
```diff
@@ -0,0 +1,13 @@
...
+<%= form_for @conn, Routes.web_auth_browser_path(Endpoint, :submit), fn f -> %>
+    <div class="form-group">
+      <%= label f, :user_code, "Enter the code displayed on your device" %>
+      <%= ViewHelpers.text_input f, :user_code %>
+      <%= error_tag @error %>
+    </div>
+
+    <button type="submit" class="btn btn-primary">Continue</button>
+
+    <br><br>
```
ericmj, line 12:
```suggestion
  <div class="form-group">
    <%= label f, :user_code, "Enter the code displayed on your device" %>
    <%= ViewHelpers.text_input f, :user_code %>
    <%= error_tag @error %>
  </div>

  <button type="submit" class="btn btn-primary">Continue</button>

  <br><br>
```

lib/hexpm_web/controllers/web_auth_browser_controller.ex
```diff
@@ -0,0 +1,57 @@
...
+        conn
+        |> put_status(:internal_server_error)
+        |> render_show(msg)
+    end
+  end
+
+  def submit(conn, _params) do
+    conn
+    |> put_status(:bad_request)
+    |> json(%{"error" => "invalid parameters"})
```
ericmj, line 34: This should not return JSON.

lib/hexpm/accounts/web_auth_request.ex
ericmj: Ping. Let me know if you need any help with this.

lib/hexpm_web/templates/web_auth_browser/success.html.eex
```diff
@@ -0,0 +1,5 @@
+<h2>Device Activation Success</h2>
+
+<p>
+    Your WebAuth request has been successfully verified.
```
ericmj, line 4:
```suggestion
  Your WebAuth request has been successfully verified.
```

lib/hexpm_web/views/api/web_auth_view.ex
```diff
@@ -0,0 +1,14 @@
...
+  use HexpmWeb, :view
+
+  def render("code.json", reponse) do
+    Map.take(reponse, [:device_code, :user_code])
+  end
+
+  def render("access.json", %{write_key: write_key, read_key: read_key}) do
+    %{
+      write_key: Routes.api_key_url(Endpoint, :show, write_key),
+      read_key: Routes.api_key_url(Endpoint, :show, read_key)
```
ericmj, line 11: The `Key.user_secret` needs to be returned here since it's generated once and never stored.

ericmj (commented):
I left a few comments, but it looks good overall. We will not merge or approve it until there is a CLI implementation as well, so that's the next step.

lib/hexpm_web/controllers/web_auth_browser_controller.ex
Benjamin-Philip: @ericmj, This is just redundant, as the submit form will always return valid parameters.

I am going to remove this.

lib/hexpm/accounts/web_auth_request.ex
Benjamin-Philip: Judging from this:

https://github.com/hexpm/hexpm/blob/bc042cacab11cdf27554bfdb06cddc6311c6dbed/lib/hexpm/accounts/audit_log.ex#L68

I should be able to manage with passing nil as audit data?

Benjamin-Philip: @ericmj,

I don't think this is an issue. The field `audit` stores the `user_agent`, which is a string. I actually was unaware of this and I think I confused every one by calling this field `audit`.

So I have a question: Should I use the browser user agent or the hex user agent? i.e. use the user agent from the submit or the access_key action?

I think we should use the latter. If we decide to use the browser `user_agent`, I will rename the `audit` field to `user_agent`.

Benjamin-Philip:
PR for a CLI implementation at https://github.com/hexpm/hex/pull/933/

lib/hexpm/accounts/web_auth_request.ex
Benjamin-Philip: @ericmj and @wojtekmach,

Ping. I am going to go ahead and use the Hex client `user_agent`.

Benjamin-Philip:
@ericmj and @wojtekmach,

Please review the Server side MVP and the Client MVP please.

Benjamin-Philip:
@ericmj, and @wojtekmach

Do review this PR and the Client side PR please.

Benjamin-Philip:
@ericmj, and @wojtekmach , could you please review? 

I would love to see this feature in production as soon as possible.

wojtekmach:
Hey, I will try to review this week. I’m looking forward to seeing this in production too, it’s gonna be great, I have some neat plans to do with it too. Your work is really appreciated.

Benjamin-Philip:
> Hey, I will try to review this week. I’m looking forward to seeing this in production too, it’s gonna be great, I have some neat plans to do with it too. Your work is really appreciated.

@wojtekmach, @ericmj Ping.

wojtekmach:
@Benjamin-Philip pong. Pinging multiple times won't change anything. I know it's extremely annoying to have to wait but I am not able to deal with this right now. Sorry.

Benjamin-Philip:
> @Benjamin-Philip pong. Pinging multiple times won't change anything. I know it's extremely annoying to have to wait but I am not able to deal with this right now. Sorry.

You're right. You are quite clearly having a busy week, and I should have been more understanding. I apologise for being impatient and pinging frequently. I'll stop pinging for now, but I hope that you don't forget about this PR.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,125 @@
...
+
+  ## Params
+
+    - `user` - The user for whom the key should be generated.
+    - `user_code` - The user code entered by that user.
+  """
+  def submit(user, user_code) do
+    request =
+      WebAuthRequest
+      |> Repo.get_by(user_code: user_code)
```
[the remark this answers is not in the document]
Benjamin-Philip: Old requests should ideally be deleted on a daily basis, but I am still looking into how schedule like that. Any suggestions?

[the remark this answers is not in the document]
ericmj: You can add the clean up as a release task here: https://github.com/hexpm/hexpm/tree/main/lib/hexpm/release_tasks and we will add a nightly job for it.

But even with the release task I would like a time filter on the query as an extra assurance in case there are any issues with the release task.

lib/hexpm/accounts/web_auth.ex
```diff
@@ -0,0 +1,132 @@
...
+        result =
+          Multi.new()
+          |> Multi.run(:write_key_gen, fn _, _ -> Keys.create(user, write_key, audit: audit) end)
+          |> Multi.run(:read_key_gen, fn _, _ -> Keys.create(user, read_key, audit: audit) end)
+          |> Multi.delete(:delete, request)
+          |> Repo.transaction()
+
+        case result do
+          {:ok, %{write_key_gen: %{key: write_key}, read_key_gen: %{key: read_key}}} ->
+            %{write_key: write_key, read_key: read_key}
```
ericmj, line 119: This should be wrapped in `{:ok, %{...}}` since we also return `{:error, ...}`.

lib/hexpm/release_tasks/remove_stale_web_auth_requests.ex
```diff
@@ -0,0 +1,30 @@
+defmodule Hexpm.ReleaseTasks.RemoveStaleWebAuthRequests do
+  require Logger
+  import Ecto.Query, only: [from: 2]
+  alias Hexpm.Accounts.WebAuthRequest
+  alias Hexpm.{Utils, Repo}
+
+  def run() do
+    # Trigger error_handler and rollbar reporting on 'hexpm eval ...'
```
ericmj, line 8: Good catch!

ericmj (commented):
This looks good overall to me. But since it is very important we get this right it will need reviews from more people. This may take some time so please bear with us.

Thank you so much for your contributions!

lib/hexpm/release_tasks/remove_stale_web_auth_requests.ex
Benjamin-Philip: Thanks!

Benjamin-Philip:
@josevalim Ping.

ericmj:
Please don't ping. We have not forgotten about your PR.

Benjamin-Philip:
Hi @ericmj, I have 2 questions:

1. Should I work on support for user supplied key creation parameters like [you mentioned](https://github.com/hexpm/hexpm/pull/1058#discussion_r760960304)

2. Can I rebase to current HEAD because it is difficult to work on other PRs to Hexpm due to different dependency versions between this branch and HEAD. Do note that we might loose/corrupt our review history.

Benjamin-Philip:
@ericmj and @josevalim, could we request someone else to review this PR as well?

This PR went out of date and had errors due to changes in main. I think we should try to merge this PR as soon as possible to prevent such errors.

Benjamin-Philip:
Also I rebased this PR to main.

Benjamin-Philip:
No worries, @ericmj.

Issue #1060, opened 2021-10-17
Duplicate undeletable user keys


lpil:
Hello!

I've got some user keys showing up twice in my user settings. If I try and delete them I get a 500 error page.

![image](https://user-images.githubusercontent.com/6134406/137639512-d08046ad-4c60-4755-9c47-490816dc65ec.png)

My username is `lpil`

Thanks,
Louis

PR #1061, opened 2021-10-19
Do not allow duplicate keys

treble37:
- Fixes https://github.com/hexpm/hexpm/issues/1060

Notes:

Testing this out locally, it looks like the web UI doesn't allow the user to generate the same key name via the dashboard (perhaps it once did?). Not sure if there's another "safe field" on the Key schema that would uniquely identify the key to delete and you'd update the form field on the view (and subsequently the controller params) with that to uniquely identify the key. Seems like the "simple fix" is to allow the user to delete the duplicate key, the app deletes the first one it finds, and the user can re-generate a new API key with the permissions they need if needed.

### DIFF
--- a/lib/hexpm/accounts/key.ex
+++ b/lib/hexpm/accounts/key.ex
@@ -169,24 +169,28 @@ defmodule Hexpm.Accounts.Key do
         s in source,
         join: k in assoc(s, :keys),
         where: not query_revoked(k),
+        where: k.name == ^name or like(k.name, ^(name <> "-%")),
         select: k.name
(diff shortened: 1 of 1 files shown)


treble37:
Added a partial unique index (seems like the unique constraint syntax above would throw an error) and reordered the other commit and its reversion for easier squashing. Can squash if you'd like...I saw `test/hexpm/accounts/key_test.exs` and it seems like `test "create unique key name", %{user: user}` ensures someone is now always creating a unique key by user...

PR #1172, opened 2022-12-12
Display latest release timestamp in "Most Downloaded"

wojtekmach:
Previously we showed _package_ timestamp which was incorrect, e.g.:

```
phoenix
8 833 678 downloads
Peace of mind from prototype to production
1.6.15 published 3156 days ago
                 ^^^^
```

### DIFF
--- a/lib/hexpm/repository/package.ex
+++ b/lib/hexpm/repository/package.ex
@@ -9,6 +9,7 @@ defmodule Hexpm.Repository.Package do
     field :name, :string
     field :docs_updated_at, :utc_datetime_usec
     field :latest_version, Hexpm.Version, virtual: true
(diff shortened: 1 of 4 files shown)


lib/hexpm/repository/package.ex
```diff
@@ -9,6 +9,7 @@ defmodule Hexpm.Repository.Package do
     field :name, :string
     field :docs_updated_at, :utc_datetime_usec
     field :latest_version, Hexpm.Version, virtual: true
```
wojtekmach, line 11: with this patch we can remove latest_version in favour of the new latest_release. which means in the other scenarios we'd fetch a bit more data from the db that we don't need but that's most likely fine. wdyt?

ericmj (approved):

PR #1196, opened 2023-05-10
Support GitHub Secret Scanning

houllette:
Changes in this PR are in order to support the [Secrets Scanning Partner Program](https://docs.github.com/en/code-security/secret-scanning/secret-scanning-partner-program) from GitHub.

**The most notable change found within is that newly generated user_secrets will now prepend `hexpm_` to the base16 encoded strings.** e.g. `hexpm_316e6532776f3368726565`

Other changes are just to support the GitHub webhooks for the Secret Scanning flow.

### DIFF
--- a/lib/hexpm/accounts/auth.ex
+++ b/lib/hexpm/accounts/auth.ex
@@ -81,7 +81,7 @@ defmodule Hexpm.Accounts.Auth do
   def gen_key() do
     :crypto.strong_rand_bytes(16)
     |> Base.encode16(case: :lower)
-    |> (fn s -> "hexpm_#{s}" end).()
+    |> String.replace_prefix("", "hexpm_")
   end
 
   defp find_email(nil, _email) do
--- a/lib/hexpm/accounts/auth.ex
+++ b/lib/hexpm/accounts/auth.ex
@@ -81,6 +81,7 @@ defmodule Hexpm.Accounts.Auth do
   def gen_key() do
     :crypto.strong_rand_bytes(16)
     |> Base.encode16(case: :lower)
+    |> (fn s -> "hexpm_#{s}" end).()
   end
 
   defp find_email(nil, _email) do
--- a/test/hexpm/accounts/keys_test.exs
+++ b/test/hexpm/accounts/keys_test.exs
@@ -32,7 +32,10 @@ defmodule Hexpm.Accounts.KeysTest do
       assert {:ok, %{key: key}} = Keys.create(user, params, audit: audit_data(user))
       assert key.name == "keyname"
       assert key.user_id == user.id
-      assert {:ok, _} = Base.decode16(key.user_secret, case: :lower)
+
+      [prefix, user_secret] = String.split(key.user_secret, "_")
+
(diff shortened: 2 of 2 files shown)


ericmj:
Hi @houllette! Are you interested in picking up this PR again?

houllette:
Hey! I would **love** to contribute to this more, but I cannot confirm when/if I have the time at the moment - if there is some desire to get this work going by someone else; by all means run with it! I'll be happy just to have been able to cast some light on the feature add 🙂

supersimple:
Note for continued work on this feature: 
image

Once we've committed to this pattern for keys, we need to email github with the pattern

supersimple:
The regexp to submit, based on the code in this PR is: `/^hexpm_[0-9a-f]{32}$/`

supersimple:
hi @houllette - I moved (and continued work) in a different branch that is committed to this repo `supersimple/github-secret-scanning`
If you want to continue work, let's use that branch since we can both contribute.
I have done my best to leave notes on the parts that are WIP, but also ask questions here if you have any.

Thanks

ericmj:
We are considering using JWTs for some tokens in the future. Are these compatible with the secret scanning program since they do not allow adding a prefix such as `hexpm_` without breaking the spec afaict?

supersimple:
> We are considering using JWTs for some tokens in the future. Are these compatible with the secret scanning program since they do not allow adding a prefix such as `hexpm_` without breaking the spec afaict?

well, yes. JWTs couldn't use a prefix. Also, one reason to use JWTs are their short TTLs, so I don't think the GH secret scanning would be of much help.

houllette:
@ericmj could you help me understand the intended usage of JWTs in the future? While they would have a shorter TTL in the event of a leakage - unfortunately they would still be usable until that point. 

I've had to go down the path of revoking JWTs before and it's not a particularly fun one, but it depends on the use case.

ericmj:
We want to use JWTs to be able to do authentication at the edge/CDN. Right now every request for a private package also hits our API which is bad for scaling reasons. A possible solution for this would be to hit the API once to generate a JWT and use that for later repository/package requests.

Users may want to move this JWT around depending on their CI setup which could leak it in output but it's probably unlikely it would be committed to a repo.

supersimple:
If our edge provider (Fastly?) can generate JWTs, my understanding is that it would be best to expire them at frequent intervals (minutes or hours.) If the common use case is to request a private package, that doesnt seem a situation where the user would need to token for more than a minute (a minute gives them time to install multiple private packages and accounts for the repo to be slow.)
If they need to re-run CI, or install from another workflow in their CI, I don't think it is a big deal to request another JWT.
As @houllette mentioned, revoking JWTs is very hard (compared to issuing new ones), so I think the better way to think about it is "what is the soonest we can expire this token?"

maennchen:
In the meantime the secret scanning spec has evolved. It now includes a 32 bit checksum and an automatic revocation endpoint for secrets found by scanners.

ericmj:
Closing in favor of: https://github.com/hexpm/hexpm/pull/1536

PR #1204, opened 2023-07-07
Increases TOTP shared secret security bits from 80 to 160

grzuy:
As per recommended by RFC 4226

From https://datatracker.ietf.org/doc/html/rfc4226#section-4:

>  R6 - The algorithm MUST use a strong shared secret.  The length of
>  the shared secret MUST be at least 128 bits.  This document
>  RECOMMENDs a shared secret length of 160 bits.

---

The current implementation causes for example for app FreeOTP to warn "Token is unsafe" when scanning the QR code.

Curiously GitHub itself also appears to have a less-than-recommended OTP secret security: https://github.com/freeotp/freeotp-android/issues/334.

---

This change only affects generation of future secrets.
Existing secrets on the database remain unchanged of course. Doing so would invalidate all users token setups.

### DIFF
--- a/lib/hexpm/accounts/tfa.ex
+++ b/lib/hexpm/accounts/tfa.ex
@@ -9,6 +9,10 @@ defmodule Hexpm.Accounts.TFA do
     embeds_many :recovery_codes, Hexpm.Accounts.RecoveryCode
   end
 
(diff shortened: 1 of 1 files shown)


ericmj (approved):

wojtekmach:
Thank you!

PR #1213, opened 2023-10-13
Update ecto dependency to work with OTP 26+.

nashby:
I'm using `Elixir 1.15.4 (compiled with Erlang/OTP 26)` locally and trying to run `mix setup` and I'm getting following error:

```
== Compilation error in file lib/ecto/query.ex ==
** (Kernel.TypespecError) lib/ecto/query.ex:430: type dynamic/0 is a built-in type and it cannot be redefined
    (elixir 1.15.4) lib/kernel/typespec.ex:961: Kernel.Typespec.compile_error/2
    (stdlib 5.0.2) lists.erl:1599: :lists.foldl_1/3
    (elixir 1.15.4) lib/kernel/typespec.ex:226: Kernel.Typespec.translate_typespecs_for_module/2
```

https://github.com/elixir-ecto/ecto/blob/master/CHANGELOG.md#v395-2023-03-22


wojtekmach:
Thank you!

PR #1234, opened 2024-02-17
Typespecs for layout_view.ex and email_view.ex

bradhanks:


lib/hexpm_web/views/layout_view.ex
```diff
@@ -1,42 +1,37 @@
...
-    if title = Map.get(assigns, :title) do
-      "#{title} | Hex"
-    else
-      "Hex"
-    end
-  end
+  @spec title(map()) :: String.t()
+  def title(assigns)
+  def title(%{"title" => title}), do: "#{title} | Hex"
+  def title(_assigns), do: "Hex"
```
wojtekmach, line 12: The semantics are not quite the same as in the previous version of the code. The old code would crash if this function gets anything other than a map. (We could of course address this concern by changing the last clause to `def title(%{}), do: "Hex"`.)

We don't and have no intention of running Dialyzer so while typespec changes don't hurt it's not like we're gonna maintain them.

So I'm not sure if in these particular cases the changes are worth it. Thank you for the PR though!

PR #1258, opened 2024-06-01
Updates CI workflow to automagically format code and recommit it (if …

supersimple:
…it changed)

@wojtekmach @ericmj WDYT about adding this?
It might save some frustration in the future to just have CI run the formatter. I also think the step called "format" now is probably better referred to as "lint" based on what it does.
If you hate it, feel free to close this PR. I wont be offended.


supersimple:
Closing this to do some debugging

PR #1276, opened 2024-10-24
Integrate Vulnerability Database

maennchen:
Integrate OSV.dev vulnerability database into hex.pm.

This PR does the following:

* Adds a new table to store vulnerabilities
* Adds an updater job, which pulls in all vulnerabilities for Hex from osv.dev
* Shows affected versions as vulnerable in the UI
* Adds the vulnerabilities to the API so that mix tasks etc. can use the information to warn the user about it.

The idea would be to follow up this change with an update to hex on rebar / mix / gleam so that it will output a warning on dependency installation / an error in the audit command. Similar to how retirement is handled right now.

localhost_4000_packages_oidcc_3 0 0 (1)

localhost_4000_packages_oidcc_versions

localhost_4000_packages_oidcc_advisories

---

## TODO

* [ ] Decide on Update Interval - See https://github.com/hexpm/hexpm/pull/1276#discussion_r1816618089
* [x] Design better UI
* [ ] Incorporate PR feedback


lib/hexpm/security_vulnerability/updater.ex
```diff
@@ -0,0 +1,141 @@
...
+  @moduledoc false
+
+  use GenServer
+
+  require Logger
+
+  alias Hexpm.Repository.Packages
+  alias Hexpm.SecurityVulnerability.Disclosures
+
+  @disclosure_download_url "https://osv-vulnerabilities.storage.googleapis.com/Hex/all.zip"
```
supersimple, line 11: Where did this URL come from? I didnt see it documented on the osv.dev website.

lib/hexpm/application.ex
```diff
@@ -1,6 +1,11 @@
 defmodule Hexpm.Application do
   use Application
 
+  case Mix.env() do
+    :test -> @jobs []
```
supersimple, line 5: this style is new. we have similar logic in the file, eg. Line 79. I'd prefer it to keep that style.

lib/hexpm/security_vulnerability/updater.ex
```diff
@@ -0,0 +1,141 @@
...
+
+  @start_opts [
+    :timeout,
+    :debug,
+    :spawn_opt,
+    name: __MODULE__,
+    hibernate_after: to_timeout(second: 1)
+  ]
+  @start_keys [:start, :timeout, :debug, :spawn_opt, :hibernate_after]
+  @init_opts [update_interval: to_timeout(minute: 30)]
```
supersimple, line 24: This pulls down the zip file every 30 minutes? IMO that is overkill.
The last vulnerability identified was yesterday, but the one before that was 6 months ago.
I'd like to stay on top of things, but that hit rate seems pretty low.

Screenshot 2024-10-25 at 6 43 59 AM

supersimple (commented):
Thanks for your contribution! I have a couple comments, but also think it's worth a larger discussion.

lib/hexpm/security_vulnerability/updater.ex
maennchen: https://google.github.io/osv.dev/data/#data-dumps

lib/hexpm/security_vulnerability/updater.ex
maennchen: I chose the interval so that a vulnerability is recognized and propagated on quickly. Since it uses the ETag to avoid reading the same file multiple times, this should be relatively inexpensive.

supersimple:
I value supply chain security, and I think a tool like this would be valuable.
A few thoughts on the implementation:
1. The database is exceedingly small for Hex (right now) and it puts me in the frame of mind that this feature should be considered as the rare exception in the code.
2. The closest analogue to this in our code now is the pwned passwords, which we actually hit their API on each request to check. I wonder - could we run this check daily perhaps on our end and then add a flag to the releases table for each affected version like we do for retirement? One benefit would be the API since we are then still working with releases and not other tables.
3. I bet we could design a better UI to alert users when a vulnerability exists. The seriousness, i think, warrants some extra thought to how we present it to users.
4. Most users will look at the hex.pm website when first installing a package, but subsequent updates are probably over CLI. As such, I think to make this more useful, we should be flagging it there as well.
5. I would like to alert the package owners when we discover a vulnerability. Ideally, they would retire the version and we could use the existing tooling to inform consumers.

maennchen:
@supersimple Thanks for your feedback.

For context, the EEF is currently in the process of becoming a CNA (CVE Numbering Authority) for all packages on Hex and a few specific GitHub repositories. This is so that we can ensure that vulnerabilities are properly disclosed and information like the `purl` is properly provided (The purl is what makes it possible to connect it to hex.pm packages.).

Currently we have a bit of a chicken / egg situation with security disclosures IMO. Good tooling to deal with vulnerabilities is not included in the standard language tool-chain since there's not many vulnerabilities published and people are not properly disclosing vulnerabilities since nobody benefits if the tooling is not used.

With upcoming legal requirements like the CRA in the EU / CISA for US government projects etc, it will get more and more important to provide tooling to handle this and other topics properly.

To make this useful in context, we need various pieces to co-exist:
* People need to properly disclose issues - https://erlef.github.io/security-wg/security_vulnerability_disclosure/
* The disclosed issues need to contain the information required for automatic processing - One of the reasons why the EEF is applying as a CNA
* Integrate the dependency vulnerability checks into the default tooling - `mix deps.get` etc, that's why this PR exists - Targeting mostly smaller projects
* Properly support SBoM in the ecosystem - Projects in planning at the moment - Targeting enterprise users
* Provide GitHub Dependency Submissions - Get Dependabot alerts etc. - https://github.com/maennchen/mix-dependency-submission - Targeting anybody on GitHub

(All of this is part of a bigger security offensive which is currently in the planning stage and also contains lots of other topics.)

---

> The database is exceedingly small for Hex (right now) and it puts me in the frame of mind that this feature should be considered as the rare exception in the code.

Unfortunately right now, yes. I hope with the projects mentioned above that this will change.

> Could we run this check daily perhaps on our end and then add a flag to the releases table for each affected version like we do for retirement?

I would recommend to have a higher check interval, but yes, that would be possible.

One thing to note is that the records are not static and might change over time. For example the affected packages could be incorrectly declared initially and then later fixed. In those case we would need to "unflag" the release.

Since we probably want to provide some extra information on the website / CLI commands, we will need to save the disclosures anyways somewhere.

> I bet we could design a better UI to alert users when a vulnerability exists. The seriousness, i think, warrants some extra thought to how we present it to users.

Completely agree. I did the UI in a few minutes without giving it much thought. This should definitely receive some attention.

Especially the CVSS Scores should be taken into account to figure out how "scary" we want to present it.

> Most users will look at the hex.pm website when first installing a package, but subsequent updates are probably over CLI. As such, I think to make this more useful, we should be flagging it there as well.

Agreed. We should have the information on the API and warn in the CLI. (Rebar3 / Mix / Gleam on Install, Update & Audit)

> I would like to alert the package owners when we discover a vulnerability. Ideally, they would retire the version and we could use the existing tooling to inform consumers.

Going forward with the CNA, we will for sure contact the maintainers anyways in the disclosure process.

However for the automatic retirement I don't necessarily agree. While I personally retire everything with a known vulnerability I maintain, this does not reflect the opinions of all maintainers. I had that discussion multiple times with many maintainers.

For example `phoenix` has never retired a version even though [CVE-2022-42975](https://osv.dev/vulnerability/GHSA-p8f7-22gq-m7j9) exists and also didn't want to do so.

Based on this, I'm inclined to leave retirement in the control of the maintainers and introducing a separate mechanism for vulnerable packages.

lib/hexpm/application.ex
maennchen: @supersimple Changed, but I removed the `Task` when empty since we would get a conflicting child `id` with multiple and instead switched it to `nil`.

ericmj:
Thank you! 💜

PR #1284, opened 2024-11-30
Cleanup the webpack config

elfenlaid:
The changes aim to clean up the webpack config. The aim here is to simplify the transition to esbuild, but also make things more explicit.

I'll go through the changes in review comments, but I'm far from being the Webpack expert. Please take my tweaks with the grain of salt 🙏

### DIFF
--- a/assets/package.json
+++ b/assets/package.json
@@ -16,7 +16,6 @@
     "phoenix_html": "file:../deps/phoenix_html",
     "phoenix_live_view": "file:../deps/phoenix_live_view",
     "phoenix": "file:../deps/phoenix",
-    "popper.js": "^1.15.0",
     "sass-loader": "^14.2.0",
     "sass": "^1.74.1",
     "style-loader": "^4.0.0",
--- a/assets/webpack.config.js
+++ b/assets/webpack.config.js
@@ -44,11 +44,6 @@ module.exports = (env, options) => ({
         test: /\.(png|woff|woff2|eot|ttf|svg)$/,
         loader: "file-loader",
         options: {outputPath: "css", publicPath: "/css"}
-      },
-      {
-        test: /bootstrap\/dist\/js\/umd\//,
-        loader: 'imports-loader',
-        options: {jQuery: 'jquery'}
       }
     ]
   },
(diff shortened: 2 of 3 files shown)


assets/package.json
```diff
@@ -16,7 +16,6 @@
     "phoenix_html": "file:../deps/phoenix_html",
     "phoenix_live_view": "file:../deps/phoenix_live_view",
     "phoenix": "file:../deps/phoenix",
-    "popper.js": "^1.15.0",
```
elfenlaid, line 19: I could be wrong, but neither the codebase, nor any of its dependencies use `popper.js`

assets/yarn.lock
```diff
@@ -2300,7 +2300,7 @@ path-type@^5.0.0:
   version "1.7.2"
 
 "phoenix_html@file:../deps/phoenix_html":
-  version "3.3.1"
```
elfenlaid, line 2303: Yarn gets the `phoenix_html` package from the Elixir `deps` directory. Currently, it's of `3.3.3` version. In other words, the change is not related to the `package.json` changes, but I thought that it would be okay to include it here.

assets/webpack.config.js
```diff
@@ -44,35 +44,14 @@ module.exports = (env, options) => ({
         test: /\.(png|woff|woff2|eot|ttf|svg)$/,
         loader: "file-loader",
         options: {outputPath: "css", publicPath: "/css"}
-      },
-      {
```
elfenlaid, line 48: Notice that webpack checks for the `/umd` directory. UMD stands for Universal Module Definition. Seems that Bootstrap 3.4.1 doesn't have these module definitions. They were added in version 4, and now in version 5 they are gone. 

Here's the screenshot of the bootstrap `dist` folder:

CleanShot 2024-11-30 at 21 15 20@2x

As of now, it doesn't affect bundled files. Though, we can remove `/umd` postfix to inject jquery here, but it seems a more reliable way of injection is to use `ProvidePlugin`. The plugin is also very close to what `esbuild` does with its injections.

assets/webpack.config.js
```diff
@@ -44,35 +44,14 @@ module.exports = (env, options) => ({
...
-        options: {jQuery: 'jquery'}
       }
     ]
   },
   plugins: [
     new MiniCssExtractPlugin({ filename: 'css/app.css' }),
     new CopyWebpackPlugin({patterns: [{ from: 'static/', to: '.' }]}),
     new webpack.ProvidePlugin({
-      $: "jquery",
       jQuery: "jquery",
```
elfenlaid, line 54: Leaving `jQuery: "jquery",` for the bootstrap jquery injection.

assets/webpack.config.js
```diff
@@ -44,35 +44,14 @@ module.exports = (env, options) => ({
...
-      "window.jQuery": "jquery",
```
elfenlaid, line 61: As far as I understand, `ProvidePlugin` implicitly imports definitions from this dictionary to bundled js files. There's only one file in the application – `app.js`, and it doesn't use any of these definitions directly.

`app.js` uses some bootstrap functionality, like button tooltips and such, but they are available through `import "bootstrap"`.

I could be wrong, but I think they can be safely removed.

assets/package.json
ericmj: Could it be used by this or is that another popper implementation?

Screenshot 2024-12-03 at 11 04 55

ericmj: It's still working so you are right.

ericmj (approved):

assets/package.json
elfenlaid: Yeah, I think it's Bootstrap `tooltip`, similar to what those copy buttons show:

CleanShot 2024-12-03 at 11 23 22@2x

ericmj:
Thank you!

elfenlaid:
Thank you for looking into it! 🙇

PR #1296, opened 2025-02-20
Allow getting release manifests without download count

f355:
Many build tools rely on checksums to ensure a remote resource didn't change unexpectedly, Hex itself does it for package tarballs. Having the ever-changing `downloads` field in the manifests makes it impossible to checksum them.

This change fixes that and allows the release manifests to be used in a hermetic manner.

### DIFF
--- a/lib/hexpm_web/controllers/api/release_controller.ex
+++ b/lib/hexpm_web/controllers/api/release_controller.ex
@@ -17,7 +17,7 @@ defmodule HexpmWeb.API.ReleaseController do
        ]
        when action in [:create, :publish, :delete]
 
-  @download_period_params ~w(day month all)
+  @download_period_params ~w(day month all none)
 
   def publish(conn, %{"body" => body} = params) do
     replace? = Map.get(params, "replace", true)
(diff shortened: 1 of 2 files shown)


ericmj:
Can you elaborate on the use case for checksumming the API responses? If it's for verifying the package it would be better to use the registry for that.

f355:
There's quite a lot to it, but I'll try to explain.

I'm currently working on an Erlang plugin for Bazel - it's a universal build system that focuses on reproducible hermetic builds and heavily relies on caching things.

When adding external hex dependencies, there's intentionally no recursive fetching, each of the dependencies needs to be specified manually, with the coordinates and the checksum. There's no problem to checksum the tarballs themselves - it works, and it works fine, but there's a caveat - when building something (a target in Bazel lingo) that depends on, say, `cowboy`, I need to also pull in cowboy's transitive dependencies (like `cowlib`) into that target's build context. I can specify that the target depends on `cowlib` as well as `cowboy`, and it would work, but it gets tedious really quickly.

So I'm using the `requirements` field from the release JSON to specify that `cowboy` depends on `cowlib`. There are other places where I can get that info, of course, like `.app`/`app.src` files, `rebar.config`, `erlang.mk`'s `Makefile`, but none of that is as reliable as the API response.

Due to the way Bazel works, not being able to checksum the API responses means not being able to cache them, so every build needs to fetch them all over again, which is unnecessary. So here we are.

I hope this makes at least a bit of sense :)

ericmj:
Thanks for the explanation. Build tools should use the registry for getting that kind of information, it is faster and more reliable than the API and it provides all the dependency data you need and everything is signed and checksummed. 

You can find the documentation here:
* https://github.com/hexpm/specifications/blob/main/registry-v2.md
* https://github.com/hexpm/specifications/blob/main/endpoints.md#repository

f355:
OK, thanks, I didn't know the registry existed.
That said, I also didn't know `metadata.config` is a part of the tarball! I'll just use that instead, problem solved.

Sorry for the noise.

PR #1311, opened 2025-07-07
Add api paramter to limit reported downloads to a date range

LostKobrakai:
For old releases the downloads map/kw list can become quite long and take a while on the api request. This allows continuous consumers* to limit to a more relevant range instead of always requesting everything.

\* I was looking at a monthly email of downloads of my personal packages.


ericmj:
Thank you!

PR #1314, opened 2025-07-14
Add Gleam usage page

lpil:


lpil:
Sorry @supersimple, I must have accidentally undone my changes in the controller before committing so it was serving the rebar3 template by mistake. I have force-pushed the correct version now, thank you.

lib/hexpm_web/templates/docs/layout.html.eex
supersimple, line 1: Could you add an external link to Tasks like the other have? Pointing to this: https://gleam.run/command-line-reference/

supersimple (changes requested):
Could you also update the link on the homepage Gleam usage widget? it points to rebar usage currently/

Thank you @lpil

lib/hexpm_web/templates/docs/layout.html.eex
lpil: Done! Thank you

wojtekmach:
This is now live, thanks!

PR #1320, opened 2025-08-24
Add support for Bluesky user handle on profiles

### LINKED ISSUE #1316: Bluesky URL on profile
It would be very nice if we could get Bluesky as a social link option on profiles.

Right now the only options are Twitter (which I no longer use) and GitHub:

Image

Image

jorams:
Fixes #1316

This adds the option for users to set their Bluesky profile on their dashboard:

image

It will then be displayed on their profile like so:

image

I also noticed the audit log fields for `UserHandles` was missing `:elixirforum` and `:slack`. I added them in a separate commit first (and reordered the list to match the schema).


ericmj:
Thank you!

PR #1322, opened 2025-09-14
Throttle logins

ericmj:


maennchen:
Allows to resolve `The package repository implements account security measures like brute force prevention (even with 2FA attempts)` in #1290 / `Authentication` / `Level 1`

PR #1334, opened 2025-10-19
Add email notifications for TFA changes

daidoji:
@ericmj @maennchen

I have a sketch of TFA enabled email change (and some other changes for the other TFA notifications) but I was actually unsure what the mitigation would be if they hadn't enabled TFA on their account?  

Email support seems like the only recourse?  Similarly with a rotation of the recovery codes?

ie) https://github.com/hexpm/hexpm/pull/1334/files#diff-d467c1a1fff0babd3aa5df3cab119351531650cbc1712206427a8f2d42ad0633R8

### DIFF
--- a/test/hexpm_web/controllers/dashboard/security_controller_test.exs
+++ b/test/hexpm_web/controllers/dashboard/security_controller_test.exs
@@ -1,5 +1,6 @@
 defmodule HexpmWeb.Dashboard.SecurityControllerTest do
   use HexpmWeb.ConnCase, async: true
+  use Bamboo.Test
 
   setup do
     %{user: insert(:user_with_tfa)}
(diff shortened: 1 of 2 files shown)


ericmj:
This looks good. I think we also need an email for `tfa_disable_app` since that effectively disables TFA also. When the templates are complete I think we can merge this.

daidoji:
@ericmj oh okay cool.  I can add that.  Honestly, I was held up looking for a way to test this but couldn't quite figure out how the email mocking was working on the controller side vs where these emails were getting sent.

I will fix formatting and add one for disable app.

ericmj:
I added the missing tests. Thank you for the PR! 💜

PR #1339, opened 2025-10-26
Bump hex_core dependency

san650:
As part of fixing this issue https://github.com/hexpm/hex/issues/1092 we need to update `hexpm` to include this fix https://github.com/hexpm/hex_core/pull/163


ericmj:
Can you add a test to make sure we don't break this in the future?

san650:
@ericmj sure, I'll work on it

ericmj:
Thank you 💜

PR #1358, opened 2025-12-13
Enable users to opt out of optional emails

Ch4s3:
This is a pretty vanilla CRUD implementation of [917](https://github.com/hexpm/hexpm/issues/917).

- New OptionalEmails module (lib/hexpm/accounts/optional_emails.ex): manages optional email preferences with validation and normalization
- Database migration: adds optional_emails JSONB field to users table
- User schema updates: adds optional_emails field and changeset functions for validation
- Email preferences UI: adds settings page in email dashboard with checkboxes to opt out of optional emails
- Organization invite email opt-out: checks user preferences before sending organization invite emails
- New API endpoint:  /dashboard/email/options to update email preferences
- Audit logging: tracks email preference changes in audit log (this may be unnecessary but I aired on the side of caution)
- Default preferences: new users get default preferences (all enabled) automatically

### DIFF
--- a/lib/hexpm/accounts/optional_emails.ex
+++ b/lib/hexpm/accounts/optional_emails.ex
@@ -12,28 +12,24 @@ defmodule Hexpm.Accounts.OptionalEmails do
     %{
       id: :organization_invite,
       title: "Organization invites",
-      description: "You’ll be notified when you are invited to an organization.",
-      optional: true
+      description: "You’ll be notified when you are invited to an organization."
     },
     %{
       id: :owner_added_to_package,
       title: "Owner added to package",
       description:
-        "You’ll be notified whenever an owner is added to a package you own or when you are added as an owner.",
-      optional: true
+        "You’ll be notified whenever an owner is added to a package you own or when you are added as an owner."
     },
     %{
       id: :owner_removed_from_package,
       title: "Owner removed from package",
       description:
-        "You’ll be notified whenever an owner is removed from a package you own or when you are removed as an owner.",
-      optional: true
+        "You’ll be notified whenever an owner is removed from a package you own or when you are removed as an owner."
     },
     %{
       id: :package_published,
       title: "Package published",
-      description: "You’ll be notified whenever a package you own is published.",
-      optional: true
+      description: "You’ll be notified whenever a package you own is published."
     }
   ]
 
@@ -68,7 +64,7 @@ defmodule Hexpm.Accounts.OptionalEmails do
   def normalize_preferences(params) when is_map(params) do
     Enum.into(@types, %{}, fn %{id: id} ->
       key = to_string(id)
-      value = params[key] || params[String.to_atom(key)]
+      value = params[key]
       {key, to_bool(value)}
     end)
   end
--- a/lib/hexpm_web/views/dashboard_view.ex
+++ b/lib/hexpm_web/views/dashboard_view.ex
@@ -302,7 +302,12 @@ defmodule HexpmWeb.DashboardView do
   defp format_email_preference_changes(changes) do
     changes
     |> Enum.map_join(", ", fn {key, value} ->
-      %{title: title} = Enum.find(OptionalEmails.list(), &(to_string(&1.id) == key))
+      title =
+        case Enum.find(OptionalEmails.list(), &(to_string(&1.id) == key)) do
+          %{title: title} -> title
+          nil -> key
+        end
+
       status = if value, do: "enabled", else: "disabled"
       "#{title}: #{status}"
     end)
--- a/lib/hexpm/accounts/user.ex
+++ b/lib/hexpm/accounts/user.ex
@@ -87,7 +87,6 @@ defmodule Hexpm.Accounts.User do
     |> validate_format(:username, @username_regex)
     |> validate_exclusion(:username, @reserved_names)
     |> unique_constraint(:username, name: "users_username_idx")
-    |> ensure_optional_email_preferences()
   end
 
   def to_organization(user, organization) do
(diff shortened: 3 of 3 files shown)


mix.lock
```diff
@@ -43,24 +43,24 @@
...
+  "phoenix": {:hex, :phoenix, "1.8.3", "49ac5e485083cb1495a905e47eb554277bdd9c65ccb4fc5100306b350151aa95", [:mix], [{:bandit, "~> 1.0", [hex: :bandit, repo: "hexpm", optional: true]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:phoenix_pubsub, "~> 2.1", [hex: :phoenix_pubsub, repo: "hexpm", optional: false]}, {:phoenix_template, "~> 1.0", [hex: :phoenix_template, repo: "hexpm", optional: false]}, {:phoenix_view, "~> 2.0", [hex: :phoenix_view, repo: "hexpm", optional: true]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:plug_cowboy, "~> 2.7", [hex: :plug_cowboy, repo: "hexpm", optional: true]}, {:plug_crypto, "~> 1.2 or ~> 2.0", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:websock_adapter, "~> 0.5.3", [hex: :websock_adapter, repo: "hexpm", optional: false]}], "hexpm", "36169f95cc2e155b78be93d9590acc3f462f1e5438db06e6248613f27c80caec"},
```
Ch4s3: happy to remove these if they should be in their own PR.

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Ch4s3: I don't use Yarn so I'm not sure if I made a mistake here.

lib/hexpm_web/templates/dashboard/email/index.html.eex
```diff
@@ -85,5 +85,42 @@
...
+                <% end %>
+              </p>
+            </li>
+          <% end %>
+        </ul>
+        <div class="panel-body">
+          <button type="submit" class="btn btn-primary">Save preferences</button>
+        </div>
+      <% end %>
+    </div>
```
Ch4s3: Tried to do the least possible here, since there may or may not be more in the future.

lib/hexpm/accounts/audit_log.ex
```diff
@@ -193,6 +193,10 @@ defmodule Hexpm.Accounts.AuditLog do
   defp extract_params("email.public", {old_email, new_email}),
     do: %{old_email: serialize(old_email), new_email: serialize(new_email)}
 
+  defp extract_params("email.options", optional_emails) when is_map(optional_emails) do
```
ericmj: We need to any new audit logs here also: https://github.com/hexpm/hexpm/blob/main/lib/hexpm_web/views/dashboard_view.ex#L50

lib/hexpm/accounts/optional_emails.ex
```diff
@@ -0,0 +1,91 @@
...
+  @moduledoc """
+  Defines optional emails that users can opt out of.
+  Example:
+   if OptionalEmails.allowed?(user, :org_invite) do
+      Emails.organization_invite(organization, user)
+      |> Mailer.deliver_later!()
+    end
+  """
+
+  @types [
```
ericmj, line 11: I think we should also add:

* Owner added to package
* Owner removed from package
* Package published

lib/hexpm/accounts/organizations.ex
```diff
@@ -174,7 +176,9 @@ defmodule Hexpm.Accounts.Organizations do
   end
 
   defp send_invite_email(organization, user) do
-    Emails.organization_invite(organization, user)
-    |> Mailer.deliver_later!()
+    if OptionalEmails.allowed?(user, :org_invite) do
```
ericmj:
```suggestion
    if OptionalEmails.allowed?(user, :organization_invite) do
```

Let's use the same key name as the email name.

lib/hexpm_web/templates/dashboard/email/index.html.eex
```diff
@@ -85,5 +85,42 @@
...
+                    value: "true",
+                    checked: Map.get(@optional_email_preferences, to_string(type.id)) %>
+                  <strong><%= type.title %></strong>
+                </label>
+              </div>
+              <p>
+                <%= if Map.get(@optional_email_preferences, to_string(type.id), true) do %>
+                  <%= type.description_enabled %>
+                <% else %>
+                  <%= type.description_disabled %>
```
ericmj: Having the description change when the checkbox status changes feels confusing, it becomes a double negative.

The text "You will not be notified when an organization you belong to adds you." with the checkbox unchecked communicates the opposite of what we want imo. I would suggest removing `description_disabled`.

yarn.lock
ericmj: We don't need this file.

lib/hexpm_web/views/dashboard_view.ex
```diff
@@ -114,6 +114,10 @@ defmodule HexpmWeb.DashboardView do
     "Set email #{params["email"]} as public email"
   end
 
+  def humanize_audit_log_info(%AuditLog{action: "email.options", params: params}) do
+    "Updated email options to #{params["optional_emails"]}"
```
ericmj: Is `params["optional_emails"]` a map? If so the string interpolation wont work

Ch4s3: this was supposed to be  "Updated email options to #{inspect(params["optional_emails"]})"

Ch4s3: rethinking this a bit, I'm going to change this to show what changed rather than dumping the whole thing on the dashboard page.

Ch4s3: image

I changed up how this gets sent to audit logs, and I think this is a better approach than what I was doing before.

Ch4s3:
@ericmj Happy New Year! Is anything else needed to merge this?

PR #1360, opened 2025-12-16
Add search suggestions with autocomplete to home page

Ch4s3:
This PR adds a LiveView-based autocomplete search on the homepage that suggests packages as users type.

#### User interface:
- Replaces the static search form on the homepage with a LiveView, `SearchSuggestionsComponent`
- Shows up to 8 suggestions in the dropdown as a users types
- Supports keyboard navigation with arrow keys and Enter to navigate to the package page
- Adds `Cmd+K`/`Ctrl+K` keyboard shortcut to focus the search input

#### Component structure:
- Implements `SearchSuggestionsLive` with debounced search and keyboard event handling
- Suggestions dropdown built as a functional component `SearchSuggestionsComponent`

#### Search implementation:
- Adds `Packages.suggest/3` with weighted ranking: exact match, prefix match, substring match, full-text description search, and recent downloads
- Uses [PostgreSQL trigram](https://www.postgresql.org/docs/current/pgtrgm.html) for name matching and GIN indexes for full-text search
- Returns highlighted descriptions with matching terms wrapped in `<strong>` tags


lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+          (CASE WHEN lower(?) LIKE ? THEN 2.0 ELSE 0 END) +
+          (CASE WHEN lower(?) LIKE ? THEN 1.0 ELSE 0 END) +
+          (LEAST(5.0, ln(1 + COALESCE(?, 0))) * 0.2) +
+          (COALESCE(
+            ts_rank_cd(
+              to_tsvector('english', regexp_replace((?->'description')::text, '/', ' ')),
+              to_tsquery('english', ?)
+            ),
+            0.0
+          ) * 0.4)
```
ericmj, line 349: How did you come up with the numbers in this weighting? Can you add some docs describing it.

lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+          recent_downloads: _,
+          latest_version: _
+        },
+        ...
+      ]
+
+      iex> Packages.suggest(repository, "")
+      []
+  """
+  def suggest(repository, term, limit \\ 8)
```
ericmj, line 239: Could we incorporate this for the normal package search also?

lib/hexpm_web/templates/page/index.html.eex
```diff
@@ -7,19 +7,7 @@
...
-              <label class="sr-only" for="search">Find packages</label>
-              <span class="input-group-btn">
-                <button type="submit" class="btn btn-search" tabindex="1">
-                  <%= icon(:heroicon, "magnifying-glass", width: 18, height: 18) %>
-                  <span class="sr-only">Search</span>
-                </button>
-              </span>
-            </div>
-          <% end %>
+          <%= Phoenix.Component.live_render(@conn, HexpmWeb.SearchSuggestionsLive, id: "home-search", session: %{"variant" => "home", "limit" => 8}, container: {:div, [style: "display: contents;"]}) %>
```
ericmj: Do we only want this on the landing page or on the package search in the header also?

assets/js/app.js
```diff
@@ -92,6 +92,26 @@ export default class App {
         e.target.value = value
       })
     }
+
+    // Global shortcut: Cmd+K / Ctrl+K focuses the search bar
```
ericmj: Is CMD+K standard for focusing the search bar? Is there any way we can communicate this to users for discoverability?

lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+        {pre, rest} = String.split_at(name, pos)
+        {mid, post} = String.split_at(rest, len)
+        pre_e = pre |> Phoenix.HTML.html_escape() |> Phoenix.HTML.safe_to_string()
+        mid_e = mid |> Phoenix.HTML.html_escape() |> Phoenix.HTML.safe_to_string()
+        post_e = post |> Phoenix.HTML.html_escape() |> Phoenix.HTML.safe_to_string()
+        pre_e <> "<strong>" <> mid_e <> "</strong>" <> post_e
+
+      :nomatch ->
+        name |> Phoenix.HTML.html_escape() |> Phoenix.HTML.safe_to_string()
+    end
```
ericmj: Can we move the HTML specifics to the component?

This piece of code feels complicated, would it make sense to escape everything first and then do a single `String.replace(text, search_term, "<strong>#{search_term}</strong>")`?

You can also mix HTML safe and unsafe strings in iodata using: https://hexdocs.pm/phoenix_html/4.3.0/Phoenix.HTML.html#t:safe/0. That should be safer than using `raw()` in the component.

lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+    |> String.downcase()
+    |> String.replace(~r/[^\w\s]/u, " ")
+    |> String.split(~r/\s+/, trim: true)
+    |> Enum.map(&String.slice(&1, 0, 50))
+    |> Enum.reject(&(&1 == ""))
+    |> Enum.map(&(&1 <> ":*"))
+    |> Enum.join(" & ")
+  end
+
+  defp package_href(repo_name, name) do
```
ericmj: HexpmWeb.ViewHelpers.package_href does this. Maybe it can be moved to the component?

lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+    search
+    |> String.replace(~r/(%|_|\\)/u, "\\\\\\1")
+  end
+
+  defp build_tsquery(search) do
+    search
+    |> String.downcase()
+    |> String.replace(~r/[^\w\s]/u, " ")
+    |> String.split(~r/\s+/, trim: true)
+    |> Enum.map(&String.slice(&1, 0, 50))
```
ericmj: Why the String.slice?

lib/hexpm/repository/packages.ex
```diff
@@ -122,4 +123,266 @@ defmodule Hexpm.Repository.Packages do
...
+      |> add_suggest_search_where(substr, tsquery)
+      |> add_suggest_order_by(pkg_part, prefix, substr, tsquery)
+      |> add_suggest_select(tsquery)
+      |> limit(^limit)
+      |> Repo.all()
+
+    package_ids =
+      Enum.map(package_results, fn {id, _name, _repo_id, _repo_name, _desc, _recent} -> id end)
+
+    versions_map =
```
ericmj: You can use `Packages.attach_latest_release` instead.

assets/js/app.js
Ch4s3: I'd say it's increasingly common, [Tailwind](https://tailwindcss.com/) does it and its an Algolia feature. I can take or leave it, this is a pretty common design idiom for it 
image

lib/hexpm_web/templates/page/index.html.eex
Ch4s3: we absolutely can, and I built this with that possibility in mind.

lib/hexpm/repository/packages.ex
Ch4s3: Yes

lib/hexpm/repository/packages.ex
Ch4s3: I honestly just made an educated guess as a first pass. Is there a way to easily get more seed data for the site locally? If I had a more representative data set I could tune it until the results felt nice subjectively. Then I could document the results I'm aiming for with it. 

This is one of the reasons the PR is in draft.


Essentially though this gives description matches a lower precedence than name or prefix matches using [ts_rank_cd](https://www.postgresql.org/docs/current/textsearch-controls.html) roughly:

Name exact: 3.0
Name prefix: 2.0
Name substring: 1.0
Downloads: up to ~1.0
Description: up to ~0.4

lib/hexpm/repository/packages.ex
Ch4s3: oh cool, I missed that.

lib/hexpm/repository/packages.ex
Ch4s3: it's an awkward way to limit input length to the underlying query because very long string cause issues with tsquery. I'll try to rewrite or better document this.

Ch4s3:
@ericmj it took me a bit to get back to this, but I think it should be ready for review now.

ericmj:
I am afraid this PR will conflict with https://github.com/hexpm/hexpm/pull/1496. Do you think this can be incorporated into 1496?

I think the extra search bar on the front page is unnecessary when we have one in the navbar.

I also see some rendering issues:
Screenshot 2026-04-28 at 16 28 28
Screenshot 2026-04-28 at 16 28 21

Ch4s3:
@ericmj I can definitely incorporate it. Should I point this PR to your branch?


I just saw that rendering issue too. Should be just a css thing.

Ch4s3:
@ericmj I'm leaving this PR open for the moment if you don't mind but just opened https://github.com/hexpm/hexpm/pull/1531 pointing at your branch. Give me a bit to get it in shape.

Ch4s3:
I'll probably close this in favor of #1531

Ch4s3:
> I'll probably close this in favor of #1531

PR #1367, opened 2026-01-25
Move scripts to admin tasks

ericmj:


test/hexpm/admin_tasks_test.exs
```diff
@@ -0,0 +1,392 @@
...
+
+  # Create admin user for AuditLogs.admin() to work
+  setup do
+    # Ensure admin user exists for audit logging
+    unless Repo.get_by(User, username: "admin") do
+      insert(:user, username: "admin")
+    end
+
+    :ok
+  end
```
wojtekmach:
```suggestion
  setup do
    insert(:user, username: "admin")
    :ok
  end
```

ericmj: I will move this to `seeds.exs` instead.

PR #1375, opened 2026-02-02
Add API endpoint to revoke OAuth tokens by hash

### LINKED ISSUE #1370: Add API action to revoke an OAuth refresh token with access to the refresh token itself
Hello!

I'm currently adding Hex OAuth support to Gleam's build tool, with help from EEF.

Gleam stores refresh tokens with encryption on disc, so if the user forgets their password we are unable to revoke tokens when the user asks to log-out or to log-in again.

Jonatan Männchen suggested having an endpoint where refresh tokens could be revoked using a hash of the token. Another option could be to have an id assigned with each token, and this token be returned.

Thanks,
Louis

ericmj:
Adds POST /api/oauth/revoke_by_hash endpoint that allows revoking OAuth refresh tokens using a SHA256 hash of the token. This enables clients that encrypt tokens (like Gleam) to revoke them even when the user has lost access to the decryption key.

Closes #1370

### DIFF
--- a/priv/repo/migrations/20260202233553_add_refresh_token_hash_to_oauth_tokens.exs
+++ b/priv/repo/migrations/20260202233553_add_refresh_token_hash_to_oauth_tokens.exs
@@ -1,0 +1,13 @@
+defmodule Hexpm.RepoBase.Migrations.AddRefreshTokenHashToOauthTokens do
+  use Ecto.Migration
+
+  def change do
+    alter table(:oauth_tokens) do
+      add :refresh_token_hash, :text
+    end
+
+    create unique_index(:oauth_tokens, [:refresh_token_hash],
+             where: "refresh_token_hash IS NOT NULL"
+           )
+  end
+end
--- a/lib/hexpm/oauth/token.ex
+++ b/lib/hexpm/oauth/token.ex
@@ -8,6 +8,7 @@ defmodule Hexpm.OAuth.Token do
   schema "oauth_tokens" do
     field :jti, :string
     field :refresh_jti, :string
+    field :refresh_token_hash, :string
     field :token_type, :string, default: "bearer"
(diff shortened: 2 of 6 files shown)


priv/repo/migrations/20260202233553_add_refresh_token_hash_to_oauth_tokens.exs
```diff
@@ -0,0 +1,13 @@
+defmodule Hexpm.RepoBase.Migrations.AddRefreshTokenHashToOauthTokens do
+  use Ecto.Migration
+
+  def change do
+    alter table(:oauth_tokens) do
+      add :refresh_token_hash, :text
+    end
+
```
maennchen, line 8: Would it make sense to prefill the value and then make it not nullable?

```elixir
    execute("""
    UPDATE oauth_tokens
    SET refresh_token_hash =
      encode(
        digest(refresh_token, 'sha256'),
        'hex'
      )
    WHERE refresh_token IS NOT NULL
      AND refresh_token_hash IS NULL
    """)

    alter table(:oauth_tokens) do
      modify :refresh_token_hash, :text, null: false
    end
```

Or possibly even

```elixir
add :refresh_token_hash, :text, generated: """
  encode(
    digest(refresh_token, 'sha256'),
    'hex'
  )
  """
```

ericmj: We can't backfill because we are not store the actual tokens in the database. `refresh_token` is an Ecto virtual field.

maennchen: @ericmj Oh yes, makes sense. Ignore me then 😅

lpil:
Thank you!!!

PR #1403, opened 2026-03-04
Add sudo mode for sensitive operations

maennchen:
- Adds sudo mode requirement for sensitive dashboard operations
- Users must re-authenticate (password or 2FA) when sudo expires
- Sudo is automatically granted on login and expires after a configurable timeout (default: 1 hour, 1 min in dev)

 ## Protected operations

- Security settings (password, 2FA, sessions)
- Email management
- API key management
- Organization settings
- OAuth device authorization

## Implementation

- `HexpmWeb.Plugs.Sudo` - plug for enforcing sudo mode
- `SudoController` - handles re-authentication via TOTP, recovery key, password or OAuth
- Rate limiting on sudo verification attempts

See individual commits.

### DIFF
--- a/lib/hexpm_web/plugs/sudo.ex
+++ b/lib/hexpm_web/plugs/sudo.ex
@@ -87,9 +87,4 @@ defmodule HexpmWeb.Plugs.Sudo do
       NaiveDateTime.utc_now() |> NaiveDateTime.to_iso8601()
     )
   end
-
-  @spec clear_sudo(Plug.Conn.t()) :: Plug.Conn.t()
-  def clear_sudo(conn) do
-    delete_session(conn, "sudo_authenticated_at")
-  end
 end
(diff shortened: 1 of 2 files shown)


ericmj:
How should deal with users that have only GitHub login, no password, and no 2FA? Is "Re-authenticate with GitHub" a good enough sudo check? It just proves an active browser session and the sudo check is supposed to be more than that.

With the new oauth device flow we are enforcing that the user has 2FA before we grant the `api:write` scope. Similiarly, should we enforce that users add 2FA for sudo to work without a password?

lib/hexpm_web/controllers/dashboard/organization_controller.ex
```diff
@@ -8,9 +8,7 @@ defmodule HexpmWeb.Dashboard.OrganizationController do
        when action in [
               :new,
               :create,
-              :show,
```
maennchen: Doesn't this lead to issues when doing any orga changes? GET does not do sudo, POST needs it and redirects away => POST body is lost.

maennchen: Yes, for example creating an API key will now redirect and when sudo is approved you end up with a 404 on `/dashboard/orgs/:org/keys` since the method is now `GET` and not `POST`.

PR #1428, opened 2026-03-17
Add README rendering on package pages via sandboxed iframe

ericmj:
Renders package READMEs from the preview service on readme.hex.pm (host-based routing in the same Phoenix app) and embeds them via sandboxed iframe on package pages. Includes HTML sanitization with Floki, HMAC-signed image proxy URL rewriting, highlight.js syntax highlighting, and iframe auto-resize via postMessage.

### Data flow

```
Package publish
  -> preview service extracts README from tarball
  -> stores files in GCS at files/{package}/{version}/README.md
  -> stores file list at file_lists/{package}-{version}.json

User views hex.pm/packages/phoenix
  -> hex.pm page embeds <iframe sandbox src="readme.hex.pm/phoenix/1.7.0">
     -> readme.hex.pm handled by hexpm Phoenix app (host-based routing)
        -> fetches file list from preview service
        -> finds README (prioritizes .md/.markdown over plain text)
        -> fetches README content from preview service
        -> renders markdown with Earmark (GFM)
        -> sanitizes HTML with Floki allowlist sanitizer
        -> rewrites relative URLs to preview service paths
        -> rewrites all image URLs to HMAC-signed img.hex.pm proxy
        -> wraps in full HTML page with highlight.js + CSP headers
     -> browser renders iframe content
     -> postMessage sends height to parent for auto-resize
     -> images load from img.hex.pm/fetch/{hmac}/{hex_encoded_url}
        -> img.hex.pm Fastly Compute service verifies HMAC
        -> fetches image from original host
        -> validates Content-Type is image/*, enforces size limit
        -> serves with cache headers and security headers
```

### Security layers

**HTML sanitization** (`HexpmWeb.Readme.Sanitizer`):
- Floki-based tag/attribute allowlist
- Strips `<script>`, `<style>`, `<form>`, `<iframe>`, `<meta>`, `<base>`, etc.
- Strips `javascript:` and `data:` URL schemes from `href`/`src`
- Strips `on*` event handlers
- Prefixes all user `id` attributes with `user-content-` (DOM clobbering protection)
- Adds `rel="nofollow noopener"` and `target="_blank"` to all links
- Only allows `style` attribute on `td`/`th` with `text-align` only
- `<input>` restricted to `type="checkbox"` only (task lists)

**iframe sandbox** (on the parent page):
- `sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"`
- No `allow-same-origin` — iframe scripts cannot access hex.pm cookies/DOM
- Defense-in-depth: even if sanitizer is bypassed, scripts are isolated

**Content Security Policy** (on readme.hex.pm responses):
- `default-src 'none'` — blocks everything by default
- `script-src 'nonce-...'` — only nonced highlight.js and height reporter run
- `style-src 'nonce-...'` — only nonced styles load
- `img-src` — images only load from img.hex.pm proxy
- `frame-ancestors` — only hex.pm can embed the page
- `base-uri 'none'` / `form-action 'none'`

**Image proxy** (img.hex.pm Fastly Compute service):
- HMAC-signed URLs prevent open proxy abuse
- HTTPS-only origin fetching
- Content-Type must be `image/*`
- 10MB size limit
- SVGs served with restrictive CSP (`default-src 'none'; style-src 'unsafe-inline'`)
- Self-referential loop protection

**Route isolation** (readme.hex.pm):
- Host-based routing: `host: "readme."` scope handles readme requests
- Catch-all route returns 404 for non-readme paths on the readme host
- Prevents readme.hex.pm from being used to access hex.pm app routes

### Loading behavior

1. Page loads with spinner in the readme area
2. If README exists: iframe sends `readme-height` postMessage → spinner removed, iframe shown
3. If no README: iframe sends `readme-not-found` postMessage → spinner removed, description fallback shown
4. Private packages: no iframe, description shown directly

### Infrastructure (hexpm-ops)

- **readme.hex.pm**: DNS A record pointing to hexpm GKE ingress (same app, host-based routing)
- **img.hex.pm**: Fastly Compute service (Rust) with TLS, DNS, configstore for HMAC secret
- **TLS certificates**: readme.hex.pm added to hexpm Let's Encrypt cert, img.hex.pm managed by Fastly
- Both staging and production environments

### DIFF
--- a/lib/hexpm_web/views/readme_view.ex
+++ b/lib/hexpm_web/views/readme_view.ex
@@ -36,8 +36,7 @@ defmodule HexpmWeb.ReadmeView do
     .readme dl dd { padding: 0 16px; margin-bottom: 16px; }
     .readme kbd { display: inline-block; padding: 3px 5px; font: 11px SFMono-Regular, Consolas, "Liberation Mono", Menlo, monospace; line-height: 10px; color: #444d56; vertical-align: middle; background-color: #fafbfc; border: 1px solid #d1d5da; border-radius: 3px; box-shadow: inset 0 -1px 0 #d1d5da; }
     .readme input[type="checkbox"] { margin-right: 0.5em; }
-    @media (prefers-color-scheme: dark) { .color-scheme-light { display: none !important; } }
-    @media (prefers-color-scheme: light) { .color-scheme-dark { display: none !important; } }
+    .color-scheme-dark { display: none !important; }
     """
   end
 


lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+
+  defp serve_readme(conn, package, release) do
+    version = to_string(release.version)
+
+    case fetch_readme(package.name, version) do
+      {:ok, filename, content} ->
+        html = render_readme(filename, content, package.name, version)
+
+        conn
+        |> put_resp_header("cache-control", "public, max-age=3600")
```
maennchen: Would it make sense to redirect from the URL without version to the URL with version and then cache the one with the version a lot longer?

lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+
+      :error ->
+        send_no_readme(conn)
+    end
+  end
+
+  defp fetch_readme(package_name, version) do
+    cdn_url = Application.fetch_env!(:hexpm, :cdn_url)
+
+    Enum.find_value(@readme_filenames, :error, fn filename ->
```
maennchen: This does quite a lot of HTTP calls. Should we use the storage adapter to check what file actually exists?

lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+    html
+    |> Sanitizer.sanitize()
+    |> URLRewriter.rewrite(package_name, version)
+    |> highlight_code_blocks()
+  end
+
+  @supported_languages %{
+    "elixir" => Makeup.Lexers.ElixirLexer,
+    "erlang" => Makeup.Lexers.ErlangLexer
+  }
```
maennchen: A lot of README contain other stuff like bash commands, YAML etc. Should we use https://github.com/elixir-makeup/makeup_syntect?

lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+
+  defp highlight_code_blocks(html) do
+    Regex.replace(
```
maennchen, line 127: We should operate on the earmark AST and not the HTML strign directly: https://github.com/team-alembic/clarity/blob/f3ecd3181feb7b24961c3fd653c0128cba8b859f/lib/clarity/components/markdown_component.ex#L51-L56

That way we can also remove the manual unescaping.

lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+  defp send_no_readme(conn) do
+    nonce = conn.assigns[:readme_csp_nonce] || ""
+
+    html = """
+    <!DOCTYPE html>
+    <html lang="en">
+    <head><meta charset="utf-8"></head>
+    <body>
+      <script nonce="#{nonce}">
+        window.parent.postMessage({type: 'readme-not-found'}, '*');
```
maennchen: It would be slightly better to supply the real origin instead of `*`.

lib/hexpm_web/controllers/readme_controller.ex
```diff
@@ -0,0 +1,216 @@
...
+  defp readme_page(conn, content_html) do
+    nonce = conn.assigns[:readme_csp_nonce] || ""
+
+    """
+    <!DOCTYPE html>
+    <html lang="en">
+    <head>
+      <meta charset="utf-8">
+      <meta name="viewport" content="width=device-width, initial-scale=1">
+      <style nonce="#{nonce}">#{readme_css()}</style>
```
maennchen: Could we do an actual phoenix layout for this and use normal `.eex` files? That way all of the interpolations should be XSS safe.

lib/hexpm_web/plugs/readme_content_security_policy.ex
```diff
@@ -0,0 +1,52 @@
...
+  import Plug.Conn
+
+  @behaviour Plug
+
+  @impl Plug
+  def init(opts), do: opts
+
+  @impl Plug
+  def call(conn, _opts) do
+    nonce = generate_nonce()
```
maennchen: Is there a reason for not using `PlugContentSecurityPolicy`? It should automatically create nonces etc.

lib/hexpm_web/readme/url_rewriter.ex
```diff
@@ -0,0 +1,116 @@
+defmodule HexpmWeb.Readme.URLRewriter do
```
maennchen, line 1: FYI: I checked local images with this package and there seems to be an issue with the logo: https://new.hex.pm/packages/crontab

lib/hexpm_web/plugs/readme_content_security_policy.ex
```diff
@@ -0,0 +1,52 @@
...
+        "frame-ancestors #{frame_ancestors}",
+        "base-uri 'none'",
+        "form-action 'none'"
+      ]
+      |> Enum.join("; ")
+
+    conn
+    |> assign(:readme_csp_nonce, nonce)
+    |> put_resp_header("content-security-policy", csp)
+    |> delete_resp_header("x-frame-options")
```
maennchen: There's an `ALLOW-FROM` directive for this header. I think that would be better than removing it.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Frame-Options#allow-from_origin

lib/hexpm_web/readme/sanitizer.ex
```diff
@@ -0,0 +1,142 @@
...
+
+      nil ->
+        attrs
+    end
+  end
+
+  defp style_to_align(attrs, _tag), do: attrs
+
+  defp sanitize_url_attr({name, value}) do
+    uri = URI.parse(String.trim(value))
```
maennchen: This is a bit dangerous. Browsers follow the WHATWG URL Standard. Elixir follows RFC 3986. This could result in treating URLs differently.

lib/hexpm_web/readme/sanitizer.ex
```diff
@@ -0,0 +1,142 @@
...
+      MapSet.member?(allowed, name) and not String.starts_with?(name, "on")
+    end)
+    |> Enum.flat_map(fn attr ->
+      sanitize_attribute(tag, attr)
+    end)
+    |> maybe_add_link_attrs(tag)
+  end
+
+  defp sanitize_attribute(_tag, {"href", _value} = attr), do: sanitize_url_attr(attr)
+  defp sanitize_attribute(_tag, {"src", _value} = attr), do: sanitize_url_attr(attr)
```
maennchen, line 88: I believe this allows me to embed a tracking pixel into my hex package page, no?

maennchen: Ah no, the rewriter should take care of that.

lib/hexpm_web/readme/sanitizer.ex
```diff
@@ -0,0 +1,142 @@
+defmodule HexpmWeb.Readme.Sanitizer do
```
maennchen, line 1: Should we consider using a tool specifically built for this? Something like https://github.com/rrrene/html_sanitize_ex

lib/hexpm_web/readme/url_rewriter.ex
```diff
@@ -0,0 +1,116 @@
...
+
+    cond do
+      uri.scheme != nil ->
+        url
+
+      String.starts_with?(url, "#") ->
+        url
+
+      true ->
+        path = String.trim_leading(url, "./")
```
maennchen, line 91: Should we use `Path.safe_relative`?

```elixir
iex(3)> Path.safe_relative("./foo/../bar/baz/./xyz")
{:ok, "bar/baz/xyz"}
```

lib/hexpm_web/readme/url_rewriter.ex
```diff
@@ -0,0 +1,116 @@
...
+    end
+  end
+
+  @doc """
+  Rewrites an image URL to go through the HMAC-signed image proxy.
+  """
+  def proxy_image_url(url) do
+    uri = URI.parse(url)
+
+    if uri.scheme in ["http", "https"] do
```
maennchen, line 106: What about `//example.com/foo.jpg` urls? Those are valid in a browser, are not relative and still need a proxy.

lib/hexpm_web/markdown_engine.ex
```diff
@@ -58,4 +64,30 @@ defmodule HexpmWeb.MarkdownEngine do
   defp only_alphanumeric(string) do
     String.replace(string, ~r"([^a-zA-Z0-9\-])", "")
   end
+
+  defp highlight_code_blocks(html) do
```
maennchen, line 71: Same comment as `highlight_code_blocks` in the readme controller.

lib/hexpm_web/readme/url_rewriter.ex
ericmj: That's because the logo is not included in the package. Images need a URL to github like this: https://github.com/elixir-ecto/ecto/blob/master/README.md?plain=1#L1 or be included in the package.

lib/hexpm_web/plugs/readme_content_security_policy.ex
ericmj: The docs says it's obsolete, I think the frame-ancestors CSP does the same thing.

lib/hexpm_web/readme/sanitizer.ex
ericmj: I couldn't get some of our requirements (id prefixing, rel/target injection, style-to-align, checkbox filtering) to work nicely with that library.

lib/hexpm_web/readme/sanitizer.ex
ericmj: Good catch, unfortunately there is no easy way to do WHATWG parsing in Elixir.

But I did some improvements: we now strip ASCII control characters from URLs before scheme validation to match WHATWG behavior, preventing bypasses like `java\\tscript:alert(1)`, and scheme matching is also case-insensitive. Combined with iframe sandbox and CSP, I hope this provides solid enough defense-in-depth.

lib/hexpm_web/plugs/readme_content_security_policy.ex
maennchen: Yes, for current browsers, the CSP is sufficient. What is even adding the header? Current phoenix does not set it anymore in `put_secure_browser_headers`.

lib/hexpm_web/markdown_engine.ex
maennchen: FYI: You can also operate on the Earmark AST and not the Floki one. The example in `clarity` did that. But this is not a security issue, I just thought it would be cleaner.

lib/hexpm_web/plugs/readme_content_security_policy.ex
ericmj: You are right, nothing is adding the header. I have removed `|> delete_resp_header("x-frame-options")`.

PR #1474, opened 2026-04-04
fix: Properly render nested GFM TODO lists

IgnacioGoldchluk:
The workaround in `HexpmWeb.Readme.TaskList` only converts top-level GFM TODO lists. Not sure if this is the right way to fix the issue, but I tested adding extra indentation in the local README for decimal and it renders properly (see images for before/after)
 
before
after


ericmj:
Thank you! 💜

PR #1512, opened 2026-04-20
Show all licenses separated by comma

### LINKED ISSUE #1500: Ability to expand or copy list of licenses
When a package has more than one license, only the first is shown and the rest are collapsed into a `div` you can hover over. This has two inconveniences: the hover tooltip covers the first license in the list, and the list is hover-only, so you can't copy it or leave the window while viewing it. The copy/leave-window case is rare, but `bcrypt_elixir` is one example where it comes up.

I suggest making the `div` clickable to expand the list, and/or adding a button to copy the licenses. Expand and copy solve slightly different problems, so ideally both, but either would help.

Let me know if I should put together a PR with this.

Image

The current implementation starts here:
https://github.com/hexpm/hexpm/blob/5c91dcc3f2396fcadbd90a81f2dea58f7264dd5c/lib/hexpm_web/components/package_layout.ex#L349

MikaelFangel:
Fixes #1500 

Changes the licenses in the package layout to show all separated by comma.

Here is an example of how the change handles many licenses: 
Screenshot 2026-04-20 at 07 03 47

### DIFF
--- a/lib/hexpm_web/components/package_layout.ex
+++ b/lib/hexpm_web/components/package_layout.ex
@@ -344,24 +344,8 @@ defmodule HexpmWeb.Components.PackageLayout do
                       </p>
                       <div class="flex items-center gap-1.5 flex-wrap">
                         <p class="text-grey-700 dark:text-grey-100 font-bold">
-                          {List.first(@licenses)}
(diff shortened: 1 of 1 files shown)


ericmj:
Thank you! 💜

PR #1513, opened 2026-04-20
fix: Use proper plurals for dependency and dependant counts

s3cur3:
This fixes the `/packages/:name` view to use the singular in tab labels when a package has 1 dependency or 1 dependant.

| Before | After |
|--------|------|
|  image |   image |


ericmj:
Thank you! 💜

PR #1531, opened 2026-04-28
Add search autocomplete with LiveView suggestions dropdown

Ch4s3:
## Summary

- Adds `SearchSuggestionsLive` — a LiveView that pushes package suggestions as the user types, with keyboard navigation (↑/↓/Enter/Escape) and ARIA listbox markup
- Adds `Packages.suggest/3` with weighted PostgreSQL ranking: exact name match → prefix → substring, boosted by recent downloads and `ts_rank_cd` description relevance; uses `ts_headline` for description snippets
- On `/packages` the existing `phx-change` live-search behaviour is **preserved**; the new dropdown only activates on pages where `live_search: false` (i.e. everywhere else)
- Nav dropdown uses Tailwind utility classes only — no inline `style=""` attributes — so it works correctly under the app's CSP `style-src` nonce policy
- SearchShortcut hook extended with **Cmd+K / Ctrl+K** in addition to `/`; skips hidden inputs so the shortcut always lands on the visible search bar
- `mark` reset added to Tailwind base so name-match `<mark>` highlights inherit text colour rather than showing browser-default yellow

## Test plan

- [ ] Typing in the nav search bar (non-packages pages) shows a dropdown with highlighted name matches and version/download info
- [ ] Typing in the nav search bar on `/packages` still triggers live package filtering (no dropdown)
- [ ] Keyboard navigation: ↑/↓ moves the active item, Enter navigates to the selected package, Escape closes the dropdown
- [ ] `/` and Cmd+K (macOS) / Ctrl+K (Windows/Linux) focus the search input from anywhere on the page
- [ ] No CSP errors in the browser console (no inline style violations)
- [ ] `mix test test/hexpm/repository/packages_suggest_test.exs` passes
- [ ] `mix test test/hexpm_web/live/search_suggestions_live_test.exs` passes


ericmj:
This is really great, thank you! Merging this to continue the work in the other PR. 💜

Ch4s3:
@ericmj you're welcome! Let me know if I can pitch in otherwise on the filtering work.

PR #1533, opened 2026-04-29
Decode `'` entity before syntax highlighting in README code blocks

williamthome:
`unescape_html/1` in the README pipeline decoded `&amp; &lt; &gt; &quot;` but missed `&#39;`. Makeup lib then received the literal entity and escaped its `&`, so the source contained `&amp;#39;`, which the browser decoded to a visible `&#39;` instead of `'`.

- Erlang atom `{'div', ...}` displays as `{&#39;div&#39;, ...}`:

image

Examples from https://hex.pm/packages/arizona.


ericmj:
Thank you! 💜

PR #1536, opened 2026-05-01
Add `hex_`-prefixed API key format with GitHub Secret Scanning support

### LINKED ISSUE #1019: GitHub secret scanning integration
GitHub has a [secret scanning feature](https://docs.github.com/en/code-security/secret-security/about-secret-scanning) that can alert users when they accidentally commit secrets like API tokens.

GitHub partners with service providers (which maintain these API tokens) to detect leaked secrets:

> Joining the secret scanning program on GitHub
> * Contact GitHub to get the process started.
> * Identify the relevant secrets you want to scan for and create regular expressions to capture them.
> * For secret matches found in public repositories, create a secret alert service which accepts webhooks from GitHub that contain the secret scanning message payload.
> * Implement signature verification in your secret alert service.
> * Implement secret revocation and user notification in your secret alert service.
> * Provide feedback for false positives (optional).

(https://docs.github.com/en/developers/overview/secret-scanning)

I think it would be beneficial for the community if Hex joins this program.

[GitHub recently changed the format of the tokens they maintain](https://github.blog/changelog/2021-03-31-authentication-token-format-updates-are-generally-available/) so if we were to participate, it'd probably make sense to first consider if we should change the format of our tokens too so they could be more easily detected. For example, instead of the token format being `~r/^[a-z0-9]{32}$/`, we prefix it with `hex_`: `~r/^hex_[a-z0-9]{32}$/`.

Thoughts?

Ch4s3:
## Add `hex_`-prefixed API keys with GitHub Secret Scanning partner integration

Closes #1019.

All newly generated API keys now carry a `hex_` prefix and a 32-bit CRC32 checksum, which satisfies [GitHub's Secret Scanning partner program](https://docs.github.com/en/enterprise-cloud@latest/code-security/tutorials/secret-scanning-partner-program) format requirements. When GitHub detects a leaked Hex API key in a public repository, it POSTs to a new endpoint and we immediately revoke the key and notify the owner by email.

Existing v1 keys continue to authenticate unchanged — no migration is required.

### Token format

`hex_` + 32 random hex chars + 8 CRC32 hex chars = **44 chars total**

The 8-char CRC32 suffix satisfies GitHub's requirement for a 32-bit checksum embedded in the token, enabling regex-based validation in their scanning pipeline. The checksum is verified on every auth request so corrupted/truncated tokens are rejected early.

- `Auth.gen_key/0` generates 16 random bytes, appends `<<:erlang.crc32(random)::32>>`, Base16-encodes the 20-byte result → 40 hex chars. `Key.gen_key/0` prepends `hex_`.
- `Auth.valid_checksum?/1` validates on lookup; invalid checksums short-circuit before the HMAC split lookup.
- A `token_format` column (`"v1"` / `"v2"`) is added to the `keys` table via migration. All new keys are `"v2"`.

### Authentication

- `Authorization: Bearer hex_*` headers are routed to key auth; other Bearer values go to OAuth.
- Raw (non-Bearer) keys have whitespace trimmed before lookup — CI/CD tools that pad the header value still work.
- `Authorization` added to `filter_parameters` in `config.exs` so it is never logged.

### GitHub Secret Scanning endpoint

`POST /api/github/secret-scanning`

- Verifies the ECDSA-P256-SHA256 signature using GitHub's published public keys (fetched from `api.github.com/meta/public_keys/secret_scanning`). Returns 403 on invalid signature or if the key fetch fails.
- Raw request body is cached before JSON parsing via the `CacheRawBody` plug so the signature check covers the original bytes.
- Returns one result object per alert token per the partner API spec:
  - `{"token_raw": "...", "token_type": "hexpm_api_token", "label": "true_positive"}` — key found and revoked
  - `{"token_raw": "...", "token_type": "hexpm_api_token", "label": "false_positive"}` — token not found (invalid checksum, unknown key, or already revoked)
- Sends a `key_leaked` email to the key owner on revocation.
- No authentication — GitHub identifies itself via the signature.

### Dashboard UI

New keys show their `hex_`-prefixed secret once in a "Key Generated Successfully" modal; the secret is not shown again after the modal is dismissed.

03-key-generated-hex-prefix

### Areas that need careful review

**`lib/hexpm/accounts/auth.ex` — `gen_key/0` and `valid_checksum?/1`**  
The HMAC is computed over the raw 40-char body (prefix stripped). If the app secret is ever rotated, all existing v2 keys become invalid in the same way v1 keys do — this is expected, but worth noting.

**`lib/hexpm_web/auth_helpers.ex` — `bearer_auth/2`**  
Routing is prefix-only: `String.starts_with?(token, "hex_")`. If a future OAuth JWT ever begins with`hex_` (I think this is super unlikely), it would be misrouted to key auth and silently fail.

**`lib/hexpm/accounts/auth.ex` — `strip_token_prefix/1`**  
Two-clause function: prefix match strips `hex_`, fallthrough passes the raw value. The fallthrough is intentional for v1 compatibility — removing it would break all legacy keys.

**`lib/hexpm/github/secret_scanning.ex` — `verify_signature/3`**  
GitHub public keys are fetched on every request (no caching). This is intentional for correctness during key rotation events, but adds one HTTP round-trip per alert batch.

### Known breakage in CI

The `Test Hex (v2.0 – main)` jobs run the `hex` client's own integration test suite, which contains `assert byte_size(key_a) == 32`. Our keys are now 44 bytes (`hex_` + 40 hex chars). *A companion PR to `hexpm/hex` is needed to update that assertion.*

### DIFF
--- a/lib/hexpm/accounts/keys.ex
+++ b/lib/hexpm/accounts/keys.ex
@@ -1,6 +1,8 @@
 defmodule Hexpm.Accounts.Keys do
   use Hexpm.Context
 
+  require Logger
+
   def all(user_or_organization) do
     Key.all(user_or_organization)
     |> Repo.all()
@@ -40,6 +42,28 @@ defmodule Hexpm.Accounts.Keys do
     else
       {:error, :not_found}
     end
+  end
+
+  @doc """
+  Looks up an active key by its `secret_first` hash, preloading the owning user
+  with emails. Used by the GitHub secret scanning revocation flow where we only
+  have the raw token, not the owner identity.
+
+  Returns `{key, user}` or `nil` if not found / already revoked.
+  """
+  def get_by_secret_first(secret_first) do
+    case Key.get_active_by_secret_first(secret_first) |> Repo.one() do
+      %Key{} = key -> {key, key.user}
+      nil -> nil
+    end
+  end
+
+  def revoke_immediately(%Key{} = key) do
+    Logger.info("Automated key revocation: key_id=#{key.id}")
+
+    key
+    |> Key.revoke()
+    |> Repo.update!()
   end
 
   def revoke_all(user_or_organization, audit: audit_data) do
--- a/lib/hexpm/accounts/keys.ex
+++ b/lib/hexpm/accounts/keys.ex
@@ -66,25 +66,34 @@ defmodule Hexpm.Accounts.Keys do
   end
 
   @doc """
-  Looks up an active key by its `secret_first` hash, preloading the owning user
-  with emails. Used by the GitHub secret scanning revocation flow where we only
-  have the raw token, not the owner identity.
+  Atomically revokes the active key matching `secret_first` and returns it
+  with the owning user (and their emails) preloaded. Used by the GitHub
+  secret scanning revocation flow where we only have the raw token, not the
+  owner identity.
 
-  Returns `{key, user}` or `nil` if not found / already revoked.
+  Returns `{:ok, key, user}` (where `user` is `nil` for organization-owned
+  keys), or `:not_found` when no active key matches.
   """
-  def get_by_secret_first(secret_first) do
-    case Key.get_active_by_secret_first(secret_first) |> Repo.one() do
-      %Key{} = key -> {key, key.user}
-      nil -> nil
+  def revoke_by_secret_first(secret_first) do
+    revoke_at = DateTime.add(DateTime.utc_now(), -1, :second)
+
+    query =
+      from(k in Key,
+        where: k.secret_first == ^secret_first,
+        where: is_nil(k.revoke_at) or k.revoke_at > fragment("NOW()"),
+        select: k,
+        update: [set: [revoke_at: ^revoke_at, updated_at: ^DateTime.utc_now()]]
+      )
+
+    case Repo.update_all(query, []) do
+      {1, [key]} ->
+        key = Repo.preload(key, user: :emails)
+        Logger.info("Automated key revocation: key_id=#{key.id}")
+        {:ok, key, key.user}
+
+      {0, _} ->
+        :not_found
     end
-  end
-
-  def revoke_immediately(%Key{} = key) do
-    Logger.info("Automated key revocation: key_id=#{key.id}")
-
-    key
-    |> Key.revoke()
-    |> Repo.update!()
   end
 
   def revoke_all(user_or_organization, audit: audit_data) do
--- a/lib/hexpm/accounts/keys.ex
+++ b/lib/hexpm/accounts/keys.ex
@@ -45,25 +45,34 @@ defmodule Hexpm.Accounts.Keys do
   end
 
   @doc """
-  Looks up an active key by its `secret_first` hash, preloading the owning user
-  with emails. Used by the GitHub secret scanning revocation flow where we only
-  have the raw token, not the owner identity.
+  Atomically revokes the active key matching `secret_first` and returns it
+  with the owning user (and their emails) preloaded. Used by the GitHub
+  secret scanning revocation flow where we only have the raw token, not the
+  owner identity.
 
-  Returns `{key, user}` or `nil` if not found / already revoked.
+  Returns `{:ok, key, user}` (where `user` is `nil` for organization-owned
+  keys), or `:not_found` when no active key matches.
   """
-  def get_by_secret_first(secret_first) do
-    case Key.get_active_by_secret_first(secret_first) |> Repo.one() do
-      %Key{} = key -> {key, key.user}
-      nil -> nil
+  def revoke_by_secret_first(secret_first) do
+    revoke_at = DateTime.add(DateTime.utc_now(), -1, :second)
+
+    query =
+      from(k in Key,
+        where: k.secret_first == ^secret_first,
+        where: is_nil(k.revoke_at) or k.revoke_at > fragment("NOW()"),
+        select: k,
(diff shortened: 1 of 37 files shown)


maennchen:
I don't think we should do rotation of existing keys. People can just create new ones.

Also make sure to consider the checksum required for key scanning: https://docs.github.com/en/enterprise-cloud@latest/code-security/tutorials/secret-scanning-partner-program#identify-your-secrets-and-create-regular-expressions

If we do this, we should also consider it for refresh and access tokens coming out of OAuth. Those can also be committed.

We should also provide a secret alert service that checks if the key was a true / false positive and revokes them: https://docs.github.com/en/enterprise-cloud@latest/code-security/tutorials/secret-scanning-partner-program#create-a-secret-alert-service

Ch4s3:
@maennchen can do.

thanks for the links, I couldn’t find one of those for some reason.

ericmj (approved):

ericmj:
@maennchen Could you sign off on this one as well?

I will fix the Hex client test errors.

lib/hexpm_web/plugs/cache_raw_body.ex
```diff
@@ -0,0 +1,40 @@
...
+  @impl Plug
+  def call(conn, _opts) do
+    if conn.request_path in paths() do
+      case Plug.Conn.read_body(conn, length: 1_000_000) do
+        {:ok, body, conn} ->
+          Plug.Conn.assign(conn, :raw_body, body)
+
+        {:more, _partial, conn} ->
+          Logger.warning("CacheRawBody: body exceeded 1 MB limit for #{conn.request_path}")
+          conn
```
maennchen: Can we make this an actual error? 413?

lib/hexpm_web/plugs/cache_raw_body.ex
```diff
@@ -0,0 +1,40 @@
...
+
+        {:error, reason} ->
+          Logger.warning("CacheRawBody: failed to read body: #{inspect(reason)}")
```
maennchen, line 29: 500?

lib/hexpm/github/secret_scanning.ex
```diff
@@ -0,0 +1,143 @@
...
+          %{"token_raw" => token, "token_type" => "hexpm_api_token", "label" => "true_positive"}
+
+        :not_found ->
+          %{"token_raw" => token, "token_type" => "hexpm_api_token", "label" => "false_positive"}
+      end
+    end)
+  end
+
+  defp revoke_token(token) do
+    raw =
```
maennchen: Should we call `Auth.strip_token_prefix` here instead?

lib/hexpm/accounts/keys.ex
```diff
@@ -42,6 +44,28 @@ defmodule Hexpm.Accounts.Keys do
     end
   end
 
+  @doc """
+  Looks up an active key by its `secret_first` hash, preloading the owning user
+  with emails. Used by the GitHub secret scanning revocation flow where we only
+  have the raw token, not the owner identity.
+
+  Returns `{key, user}` or `nil` if not found / already revoked.
+  """
```
maennchen, line 76: Should we make this atomic?

```suggestion
  def revoke_by_secret_first(secret_first) do
    revoke_at = DateTime.add(DateTime.utc_now(), -1, :second)

    query =
      from(k in Key,
        where: k.secret_first == ^secret_first,
        where: is_nil(k.revoke_at) or k.revoke_at > fragment("NOW()"),
        left_join: u in assoc(k, :user),
        left_join: e in assoc(u, :emails),
        preload: [user: {u, emails: e}],
        select: k,
        update: [set: [revoke_at: ^revoke_at, updated_at: ^DateTime.utc_now()]]
      )

    case Repo.update_all(query, []) do
      {1, [key]} ->
        Logger.info("Automated key revocation: key_id=#{key.id}")
        {:ok, key, key.user}

      {0, _} ->
        :not_found
    end
  end
```

maennchen (changes requested):
Left a few comments. I'm not sure how we can test this end to end. I assume we should deploy it and contact secret-scanning@github.com. They will most likely let us know on how to test it end to end.

https://docs.github.com/en/code-security/tutorials/secret-scanning-partner-program#joining-the-secret-scanning-program-on-github

maennchen (approved):

maennchen:
@ericmj Should we only roll this out to staging? If we messed something up and need to change something, it would be better if users did not already use new tokens.

ericmj:
> Should we only roll this out to staging? If we messed something up and need to change something, it would be better if users did not already use new tokens.

That's a good idea. I will deploy it to staging and email github.

ericmj:
This is currently live on staging.hex.pm and I have contacted GitHub according to the instructions here: https://docs.github.com/en/code-security/tutorials/secret-scanning-partner-program#joining-the-secret-scanning-program-on-github.

Ch4s3:
I’ll try to address the feedback ASAP, I’m a little under the weather so it may take a few days.

maennchen:
@Ch4s3 At this point, we could use a review from you. Otherwise the change should be done.

Ch4s3:
@maennchen sorry, I misread the prior comment. The commits from @ericmj looks great, its always cool to see the gaps someone who's lived with the code base for a long time can find. I'm good with moving forward. How else can I help here?

Ch4s3:
Do you guys need anything else fro me on this?

ericmj:
I think we are good for now. Will update here when we get a response from GitHub. Thank you!

PR #1538, opened 2026-05-02
Display optional dependencies badge if at least one is present

fellipessanha:
Optional badge was not being displayed on dependency pages:

image

This fixes that by shortening the second column if the third one is present:

image

### DIFF
--- a/lib/hexpm_web/templates/package/dependencies.html.heex
+++ b/lib/hexpm_web/templates/package/dependencies.html.heex
@@ -55,6 +55,7 @@ requirements_count = length(requirements) %>
                 <td class="px-5 py-3.5">
                   <a
                     href={ViewHelpers.path_for_package(req.repository, req.name)}
+                    title={if req.repository != "hexpm", do: "#{req.repository}/#{req.name}", else: req.name}
                     class="font-semibold text-grey-900 dark:text-grey-100 hover:text-purple-600 dark:hover:text-primary-300 transition-colors font-mono"
                   >
                     <%= if req.repository != "hexpm" do %>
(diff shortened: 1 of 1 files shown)


ericmj:
Can you include a screenshot of the mobile view. Try to check how it looks with longer package names and requirements.

fellipessanha:
Thanks a lot for your quick feedback, @ericmj!

This is how it looks like on the more restrictive default dimensions I could find, the Samsung Galaxy S8+:

Optional tag on S8+ dimensions as per commit e27897557aac5aa7cc3cf494020deed47dbcc01c
image

<details>
<summary>
this is how actual hex.pm looks right now, for comparison:
</summary>
image
</details>

The optional badge get a little cut off, as does the package name, latter seems to be a consequence of the `2/5` current width.

Another solution that I thought of was to merge the optional and requirements columns into a single one, and resizing the columns so the package names won't get cut so easily. I implemented it over at 17ac62c09aabe281b5f40e322ab5f3ccd80173f4, here are some prints so you can see how it looks like:
<details>
<summary>
On desktop browser:
</summary>
image
</details>

<details>
<summary>
On S8+ dimensions:
</summary>
image
</details>

I think that could be improved if additional tags may be implemented for `dev` dependencies, for instance, and I liked that it solved the current not-whole-name-being-displayed problem.

Which solution do you prefer? I'd love to contribute some more to this project.

ericmj:
Merging the two columns looks like a good idea. I also like the line break between requirement and badge on mobile view, it's a good compromise for the available space.

Maybe we should also add a title text for package name and version requirement so we can see the full text if it gets cut off by column width?

ericmj (approved):

fellipessanha:
Sorry for the delayed response, I forgot to push the changes in the weekend.

Added the title options in 26951c6fd6731054126770b97dc2ad03b551a27c

I hope I can contribute again soon, feel free to ask for help.

ericmj:
Looks great! Thank you!

PR #1549, opened 2026-05-07
Serialize Security Advisories into Repo Files

maennchen:
https://github.com/hexpm/specifications/pull/50

### DIFF
--- a/lib/hexpm/repository/registry_builder.ex
+++ b/lib/hexpm/repository/registry_builder.ex
@@ -191,10 +191,10 @@ defmodule Hexpm.Repository.RegistryBuilder do
 
   defp build_names(repository, packages) do
     packages =
-      Enum.map(packages, fn {name, {updated_at, _versions}} ->
+      Enum.map(packages, fn {name, {updated_at, _versions, _advisories}} ->
         # Currently using Package.updated_at, would be more accurate to use
         # a timestamp that is only updated when the registry is updated by:
-        # publish, revert, or retire
+        # publish, revert, retire, or new advisory
         {seconds, nanos} = to_unix_nano(updated_at)
 
         %{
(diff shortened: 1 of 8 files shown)


lib/hexpm/repository/registry_builder.ex
```diff
@@ -415,32 +479,52 @@ defmodule Hexpm.Repository.RegistryBuilder do
...
+           {p.name, p.updated_at,
+            fragment(
+              "coalesce(json_agg(json_build_object('id', ?, 'summary', ?, 'cvss_rating', ?, 'cvss_score', ?)) FILTER (WHERE ? IS NOT NULL), '[]')",
+              a.id,
+              a.summary,
+              a.cvss_rating,
+              a.cvss_score,
+              a.id
+            )}}
+      )
```
ericmj, line 500: We should add an `ORDER BY` here to make it deterministic.

lib/hexpm/repository/registry_builder.ex
```diff
@@ -191,7 +191,7 @@ defmodule Hexpm.Repository.RegistryBuilder do
 
   defp build_names(repository, packages) do
     packages =
-      Enum.map(packages, fn {name, {updated_at, _versions}} ->
+      Enum.map(packages, fn {name, {updated_at, _versions, _advisories}} ->
         # Currently using Package.updated_at, would be more accurate to use
         # a timestamp that is only updated when the registry is updated by:
         # publish, revert, or retire
```
ericmj:
```suggestion
        # publish, revert, retire, or new advisory 
```

ericmj (commented):
We are missing the trigger for registry rebuilding when we get an advisory update.

maennchen:
@ericmj True, that should be in now.

Careful, it only does changed advisories => Once a full rebuild is needed.

lib/hexpm/security/advisories.ex
```diff
@@ -273,6 +279,46 @@ defmodule Hexpm.Security.Advisories do
     end)
   end
 
+  defp rebuild_package_registries(multi) do
+    Multi.run(multi, :rebuild_package_registries, fn
```
maennchen, line 283: Also needs to consider the deleted entries from `reconcile_advisories`

ericmj (approved):

ericmj:
@maennchen Is this ready to merge?

maennchen:
@ericmj I think so, yes

PR #1560, opened 2026-05-12
Add package ownership management page

### LINKED ISSUE #1541: package ownership management
As part of https://github.com/hexpm/hexpm/issues/1290 there needs to be a UI for managing package owners so that I can be done securely from the web. This should include a workflow for inviting users from outside of Hex to become Hex users and package managers.

Ch4s3:
Introduces the owner management page at `/packages/:name/owners`, where full owners can add, remove, and change the role of co-owners. This is the same functionality the CLI provides, now available on the web. Closes [#1541](https://github.com/hexpm/hexpm/issues/1541)

## What's in this PR

- New page at `/packages/:name/owners` (visible only to full owners via the "Owners" tab and a "Manage" link in the sidebar)
- **Add owner** — plain username + level form, direct add (no invite flow)
- **Change role** — maintainer ↔ full owner via a select dropdown
- **Remove owner** — with a confirmation prompt
- All mutating actions (add/change/remove) are gated behind **sudo mode** (`HexpmWeb.Plugs.Sudo`)
- Audit logs for `owner.add` and `owner.remove`
- Notification emails to existing owners when someone is added or removed (existing behaviour via `Owners.add/4` and `Owners.remove/3`)

## What's NOT in this PR (scoped out per feedback)

- Invite/accept/decline email workflow
- Email-only invites for users without Hex accounts
- Pending invitations display on the profile dashboard


Ch4s3:
Is there a preference here for `heex` files vs `~H` in the module? Using a heex file felt like it fit into a classic phoenix app better, but I usually prefer embedded sigils in functions.

ericmj:
I agree this PR is too large right now. The goal should be to copy the package ownership management functionality from the CLI to the website so that such management is easier with MFA.

Adding package ownership invites with an approve or disapprove workflow via emails is nice and something that I think we should eventually add. On the other hand, no one has asked for this yet, so the current notification-only emails seem to be enough for now.

Additionally, adding account creation email invites for users that currently don't have emails could be nice, but I'm also hesitant about sending emails to users that don't have accounts without a captcha gate.

If you scope down the current PR to not do the email invites  we could get this merged quicker.

ericmj:
Another note: we should add sudo mode for adding/removing owners.

Ch4s3:
I'll scope it down.

Ch4s3:
ok @ericmj I think its in much better shape now.

Ch4s3:
@ericmj this all looks good to me. The modal is a good idea.

PR #1563, opened 2026-05-13
Group security advisories by aliases

### LINKED ISSUE #1559: Group Security Advisories by Aliases
Join Advisories by the `aliases` field. For example `phoenix` shows both GHSA-628h-q48j-jr6q and EEF-CVE-2026-32689 that refer to the same CVE.

For display purposes, we should deduplicate them.

For API, we can leave it as is and leave deduplication up to the consumer.

ericmj:
Closes https://github.com/hexpm/hexpm/issues/1559.

### DIFF
--- a/lib/hexpm/security/advisories.ex
+++ b/lib/hexpm/security/advisories.ex
@@ -17,6 +17,15 @@ defmodule Hexpm.Security.Advisories do
     |> Repo.preload([:references, :affected_versions])
   end
 
+  def group_for_display(advisories) when is_list(advisories) do
+    groups = Enum.group_by(advisories, &display_group_key/1)
+
+    advisories
+    |> Enum.map(&display_group_key/1)
+    |> Enum.uniq()
+    |> Enum.map(fn key -> merge_display_group(Map.fetch!(groups, key)) end)
+  end
+
   def upsert(records, package_ids) do
     Multi.new()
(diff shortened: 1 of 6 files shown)


lib/hexpm/security/advisories.ex
```diff
@@ -114,6 +123,87 @@ defmodule Hexpm.Security.Advisories do
...
+      max_datetime(Map.get(advisory, field), acc)
+    end)
+  end
+
+  defp max_datetime(nil, acc), do: acc
+  defp max_datetime(datetime, nil), do: datetime
+
+  defp max_datetime(datetime, acc) do
+    if DateTime.compare(datetime, acc) == :gt, do: datetime, else: acc
+  end
```
maennchen:
```suggestion
  defp min_datetime_field(advisories, field) do
    advisories |> Enum.map(&Map.get(&1, field)) |> Enum.reject(&is_nil/1) |> Enum.min(DateTime, fn -> nil end)
  end

  defp max_datetime_field(advisories, field) do
    advisories |> Enum.map(&Map.get(&1, field)) |> Enum.reject(&is_nil/1) |> Enum.max(DateTime, fn -> nil end)
  end
```

maennchen (changes requested):

maennchen (approved):

PR #1570, opened 2026-05-13
Submit search form natively on Enter to skip LiveView roundtrip

maennchen:
## Summary

  - When the user presses Enter in the search bar with no autocomplete suggestion active, the form is now submitted natively (bypassing the LiveView roundtrip), giving instant navigation to the packages page.
  - When a suggestion is highlighted via arrow keys (`aria-activedescendant` is set), LiveView still handles the submit to navigate to that package.

To test the difference, add `liveSocket.enableLatencySim(500)` in `assets/js/app.js` after `liveSocket.connect()`. With the delay active, pressing Enter with no suggestion highlighted should still navigate instantly, while selecting a suggestion via arrow keys will be visibly delayed.

### DIFF
--- a/assets/js/hooks/search_shortcut.js
+++ b/assets/js/hooks/search_shortcut.js
@@ -5,6 +5,10 @@
  * Ignores input when typing in form fields.
  * Skips hidden inputs (offsetParent === null).
  *
+ * Also handles Enter key on the search input: when no autocomplete suggestion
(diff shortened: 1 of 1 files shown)


Ch4s3:
This is a nice upgrade!

ericmj:
Thank you! 💜

PR #1571, opened 2026-05-14
Improve advisory dedup display: alias links and reference type merging

maennchen:
When grouping advisories for display, alias IDs that correspond to a separate advisory in the group now link to osv.dev instead of rendering as plain text. References with the same URL from different sources are merged into a single entry with all types accumulated, rather than being deduplicated by (type, url) pair.

### DIFF
--- a/lib/hexpm/security/advisories.ex
+++ b/lib/hexpm/security/advisories.ex
@@ -140,10 +140,19 @@ defmodule Hexpm.Security.Advisories do
   end
 
   defp display_aliases(primary, advisories) do
+    advisory_ids = MapSet.new(advisories, & &1.id)
+
     advisories
     |> Enum.flat_map(&display_identifiers/1)
     |> Enum.uniq()
     |> Enum.reject(&(&1 == primary.id))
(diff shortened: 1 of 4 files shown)


lib/hexpm/security/advisories.ex
```diff
@@ -179,7 +188,11 @@ defmodule Hexpm.Security.Advisories do
   defp uniq_references(advisories) do
     advisories
     |> Enum.flat_map(&loaded_assoc(&1.references))
-    |> Enum.uniq_by(&{&1.type, &1.url})
+    |> Enum.group_by(& &1.url)
+    |> Enum.map(fn {url, refs} ->
+      types = refs |> Enum.map(& &1.type) |> Enum.uniq()
+      %{url: url, types: types}
```
ericmj, line 194: Are we using `types` anywhere?

maennchen: Not right now. But it felt cleaner than throwing it away. We might want to show a pill with the type or something like that at some point.

ericmj:
Thank you! 💜

PR #1572, opened 2026-05-14
Sort typeahead suggestions by downloads within each match tier

### LINKED ISSUE #1567: New search typeahead sorting
Should the new search typeahead sort by downloads?

These are correct results, but I would maybe expect ecto to be the top result. Just based on test match score everything starting with `ect` will match equally.

Image

Ch4s3:
### Fixes #1567.

My first implementation of the search typeahead was not sorting results by downloads. Searching for `ect` returned `ecto`, `ecto_sql`, `ecto_commons`, etc. in effectively arbitrary order because the scoring formula capped the download contribution too aggressively:

```sql
(LEAST(5.0, ln(1 + COALESCE(downloads, 0))) * 0.2)
```

`e^5 ≈ 148`, so any package with more than ~150 downloads receives the maximum score of `1.0`. Since almost every published package exceeds that threshold, all candidates in the same name-match tier (exact/prefix/substring) received identical download scores and fell back to alphabetical ordering.

I should have done the math.

### Fix

Remove the `LEAST(5.0, …)` cap and reduce the multiplier to `0.1`:

```sql
(ln(1 + COALESCE(downloads, 0)) * 0.1)
```

This lets the logarithm scale continuously with actual download counts, so `ecto` (millions of weekly downloads) naturally surfaces above `ecto_commons` (thousands of weekly downloads) within the prefix-match tier.

The `0.1` multiplier keeps download weight below the `1.0`-point gap between tiers (exact → prefix → substring), so name-match quality still takes precedence — an exact match like `ecto` will not be buried by a hugely popular prefix match, unless the exact-match package has essentially zero downloads.

### DIFF
--- a/lib/hexpm/repository/packages.ex
+++ b/lib/hexpm/repository/packages.ex
@@ -241,8 +241,12 @@ defmodule Hexpm.Repository.Packages do
   def suggest(_repository, "", _limit), do: []
 
   def suggest(repository, term, limit) when is_binary(term) do
-    term = String.trim(term)
-    do_suggest(repository, term, limit)
+    term
+    |> String.trim()
+    |> case do
+      term when byte_size(term) < 3 -> []
+      term -> do_suggest(repository, term, limit)
+    end
   end
 
   defp do_suggest(_repository, "", _limit), do: []
@@ -260,7 +264,7 @@ defmodule Hexpm.Repository.Packages do
     |> where([p], p.repository_id == ^repository.id)
     |> add_suggest_joins()
     |> add_suggest_search_where(substr, tsquery)
-    |> add_suggest_order_by(pkg_part, prefix, substr, tsquery)
+    |> add_suggest_order_by(pkg_part, prefix, substr)
     |> add_suggest_select(tsquery)
     |> limit(^limit)
     |> Repo.all()
@@ -309,7 +313,7 @@ defmodule Hexpm.Repository.Packages do
     where(
       query,
       [p],
-      fragment("lower(?) LIKE ?", p.name, ^substr) or
+      like(p.name, ^substr) or
         fragment(
           "to_tsvector('english', regexp_replace((?->'description')::text, '/', ' ')) @@ to_tsquery('english', ?)",
           p.meta,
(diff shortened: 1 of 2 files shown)


lib/hexpm/repository/packages.ex
ericmj: I have two questions about this function that I missed in my initial review:
1. This is currently not using an index right? I think that could be an issue for a query that is triggered by key presses.
2. If we think this ordering is more useful than the default ordering on https://hex.pm/packages then we should it there too.

ericmj: I checked the query insights and this query is indeed a lot more expensive than our other queries, so it needs improvement:

Image

Ch4s3: I'll run it through `EXPLAIN ANALYZE` and get it under control. I may also add a bit more debounce.

I assume it hits the `packages_description_text` index, but that may be a bad assumption.

Ch4s3: Ok, it looks to be using a `seq scan`, which didn't jump out at me locally because I only have ~120` packages locally. I'll get it fixed in this PR.

Ch4s3: Ok, I removed the `"lower(?) LIKE ?"` in the `where`. It seems like that was causing it to miss the indexes, not the order_by.

  | Original (lower(name) LIKE) | Fixed (name ILIKE)
-- | -- | --
Plan | Seq Scan on packages | BitmapOr index scan
Indexes hit | none | packages_name_trgm + packages_description_text
Dev execution | 7ms | 0.27ms 

*on 160 packages


Is there a good way to get more data locally?

Ch4s3: Ok, @ericmj if I drop ts_rank_cd and lean on downloads it probably saves ~25-50ms per query with a slight hit to search relevance. I'm fine removing it.

The big savings come from fixing the issue above, but tell me which you prefer and I'll do it. Its a quick change.

lib/hexpm/repository/packages.ex
```diff
@@ -309,7 +309,7 @@ defmodule Hexpm.Repository.Packages do
     where(
       query,
       [p],
-      fragment("lower(?) LIKE ?", p.name, ^substr) or
+      ilike(p.name, ^substr) or
```
ericmj: All package names are lower case, but the query planner doesn't know that. This has the same issue as the previous query. It should be:

```suggestion
      like(p.name, ^substr) or
```

lib/hexpm/repository/packages.ex
```diff
@@ -320,22 +320,25 @@ defmodule Hexpm.Repository.Packages do
...
       desc:
         fragment(
           """
           (CASE
-            WHEN lower(?) = ?        THEN 3.0
-            WHEN lower(?) LIKE ?     THEN 2.0
-            WHEN lower(?) LIKE ?     THEN 1.0
+            WHEN ? ILIKE ?           THEN 3.0
+            WHEN ? ILIKE ?           THEN 2.0
+            WHEN ? ILIKE ?           THEN 1.0
```
ericmj:
```suggestion
            WHEN ? LIKE ?           THEN 3.0
            WHEN ? LIKE ?           THEN 2.0
            WHEN ? LIKE ?           THEN 1.0
```

lib/hexpm/repository/packages.ex
ericmj: Lets drop `ts_rank_cd` and lets not return search results until the user has typed at least 3 characters so we hit the trigram index.

I am still a bit concerned about not using an index for ORDER BY, but this should fix the larger issue and we can monitor the production stats to check if we need an ORDER BY index.

Ch4s3: should be good now. I'd like to use an index but the subquery reaches over to package_downloads on a smaller subset of data. It would either require a much more complicated query or denormalization to hit an index here, unless I've missed something (which is entirely possible).

ericmj:
Thank you! 💜

PR #1578, opened 2026-05-18
Emit published_at in registry packages

ericmj:
Add per-release publication time to the v2 registry package resource. The registry builder now selects `Release.inserted_at` and writes it as a `{seconds, nanos}` Timestamp map under each release's `published_at` field in `/packages/NAME`. Matches hex_core PR #187 which defined the Timestamp message in the proto schema.

Clients use this to compute release age and apply policies such as a configurable release-age cooldown during dependency resolution.


mix.exs
```diff
@@ -95,11 +95,14 @@ defmodule Hexpm.MixProject do
...
+  defp hex_core_dep() do
     if path = System.get_env("HEX_CORE_PATH") do
-      [path: path]
+      {:hex_core, "~> 0.16", path: path}
     else
-      []
+      # Pinned to the merge commit of hex_core PR #187, which added the
+      # Timestamp published_at field. Revert to a hex.pm release once
+      # hex_core cuts one carrying the field.
+      {:hex_core, github: "hexpm/hex_core", ref: "5882d22"}
```
ericmj: Change to hex dependency before merging.

lib/hexpm/repository/registry_builder.ex
```diff
@@ -525,6 +529,7 @@ defmodule Hexpm.Repository.RegistryBuilder do
         outer_checksum: r.outer_checksum,
         build_tools: fragment("?->'build_tools'", r.meta),
         retirement: r.retirement,
+        inserted_at: r.inserted_at,
```
ericmj, line 532: `inserted_at` is technically not the same as `published_at`, but for public packages' non-initial releases the window is only 1 hour which are the only relevant packages for the threat model, so I think that's okay.

Issue #1327, opened 2025-10-07
Notify maintainer via email for critical account security changes


daidoji:
I'm attempting to help with #1290 for the feature

> The package repository notifies maintainers via email for critical account security changes, such as password changes or disabling multi-factor authentication.

I've opened this issue to track the work on this particular item.  As its my first foray into this code base I think I've located where these types of changes are occurring in state, in action, and where they're put into the audit log.  I do have questions about:

- where this notification service should be most appropriate?
- Best way to send an email in the current project.
- Where exactly to hook in?  It seems like in action would be the simplest but I could watch the audit log.
- Are there any scope issues with this feature?
  - right now its just TFA disable and password resets but maybe we should include recovery codes used or other events that may have security concerns.

# Successful implementation
Every time one of the important security events triggers, an email is sent to the user or maintainers emails of that security event.

maennchen:
> Best way to send an email in the current project.

Check https://github.com/hexpm/hexpm/blob/main/lib/hexpm/emails/emails.ex and how it is used.

> Are there any scope issues with this feature?

I see multiple things in the settings that are security relevant. (Did not check which ones do already trigger emails.)

* Password Change
* 2FA Enable / Disable / Setup new app
* Generate new recovery codes
* Email Address Adding / Change Primary
* Add / Revoke API Key

> Where exactly to hook in? It seems like in action would be the simplest but I could watch the audit log.

Right now they are triggered where the change happens. For example for PW change:

https://github.com/hexpm/hexpm/blob/4eb659c2332d54ec204c7e67bd4d866fb45bb94c/lib/hexpm/accounts/users.ex#L171

ericmj:
Thanks for picking this up.

> I see multiple things in the settings that are security relevant. (Did not check which ones do already trigger emails.)
> 
> * Password Change
> * 2FA Enable / Disable / Setup new app
> * Generate new recovery codes
> * Email Address Adding / Change Primary
> * Add / Revoke API Key

This looks like a comprehensive list, I would also add when new OAuth tokens are created. But I would say revoking API key is not necessary to send an email for?

ericmj:
> Is there an issue / docs somewhere about the new OAuth implementation? What is the use case for it?

[...]

PR #1579, opened 2026-05-18
Add security notification emails for email and API key events

### LINKED ISSUE #1290: Track "Principles for Package Repository Security" Current State

Ch4s3:
Closes #1290 (remaining items from [this comment](https://github.com/hexpm/hexpm/issues/1290#issuecomment-4481245409)) and #1327.

## Summary

- Notifies the existing primary address when a new email is added to an account
- Notifies both old and new primary addresses when the primary email changes
- Notifies the user (or all org admins) when an API key is created
- Notifies the user (or all org admins) when a single API key is revoked
- Notifies the user (or all org admins) with a single consolidated email when all API keys are revoked at once

All notifications also write a standalone audit log entry (`security_notification.*`) with the same request context as the originating action.

## Architecture

A new `Hexpm.Accounts.SecurityNotifications` module is the single dispatch point for all these emails + audit entries. Context functions (`users.ex`, `keys.ex`) call into it after a successful transaction rather than directly into `Hexpm.Emails`.

### DIFF
--- a/lib/hexpm/emails/emails.ex
+++ b/lib/hexpm/emails/emails.ex
@@ -2,7 +2,7 @@ defmodule Hexpm.Emails do
   use Bamboo.Phoenix, view: HexpmWeb.EmailView
   import Bamboo.Email
   import Bamboo.SendGridHelper, only: [with_click_tracking: 2]
-  alias Hexpm.Accounts.{Email, User}
+  alias Hexpm.Accounts.{Email, Organization, User}
 
   def owner_added(package, owners, owner) do
     email()
@@ -82,6 +82,51 @@ defmodule Hexpm.Emails do
     |> render(:tfa_recovery_rotated)
   end
 
+  def email_added(user, new_email) do
+    email()
+    |> email_to(user)
+    |> subject("Hex.pm - A new email address was added to your account")
+    |> assign(:username, user.username)
+    |> assign(:new_email, new_email.email)
+    |> render(:email_added)
+  end
+
+  def primary_email_changed(user, old_addr, new_addr) do
+    email()
+    |> email_to(old_addr)
+    |> subject("Hex.pm - Your primary email address has changed")
+    |> assign(:username, user.username)
+    |> assign(:old_addr, old_addr)
+    |> assign(:new_addr, new_addr)
+    |> render(:primary_email_changed)
+  end
+
+  def api_key_created(user_or_org, key) do
+    email()
+    |> email_to(user_or_org)
+    |> subject("Hex.pm - A new API key was created on your account")
+    |> assign(:username, display_name(user_or_org))
+    |> assign(:key_name, key.name)
+    |> render(:api_key_created)
+  end
+
+  def api_key_revoked(user_or_org, key) do
+    email()
+    |> email_to(user_or_org)
+    |> subject("Hex.pm - An API key was revoked on your account")
+    |> assign(:username, display_name(user_or_org))
+    |> assign(:key_name, key.name)
+    |> render(:api_key_revoked)
+  end
+
+  def api_keys_all_revoked(user_or_org) do
+    email()
+    |> email_to(user_or_org)
+    |> subject("Hex.pm - All API keys have been revoked on your account")
+    |> assign(:username, display_name(user_or_org))
+    |> render(:api_keys_all_revoked)
+  end
+
   def typosquat_candidates(candidates, threshold) do
     email()
     |> email_to(Application.get_env(:hexpm, :support_email))
@@ -146,12 +191,14 @@ defmodule Hexpm.Emails do
       to
       |> List.wrap()
       |> Enum.flat_map(&expand_organization/1)
+      |> Enum.reject(&is_nil(recipient_email(&1)))
       |> Enum.sort_by(&recipient_email/1)
       |> Enum.uniq_by(&recipient_email/1)
 
     to(email, to)
   end
 
+  defp recipient_email(nil), do: nil
   defp recipient_email(email) when is_binary(email), do: email
   defp recipient_email(%Email{email: email}), do: email
   defp recipient_email(%User{} = user), do: User.email(user, :primary)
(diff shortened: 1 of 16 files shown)


Ch4s3:
> - Notifies the existing primary address when a new email is added to an account

@ericmj did I misunderstand what you were asking for here? It seems like there are a few valid ways this could be done but it makes sense to me to email the old address.

ericmj:
> did I misunderstand what you were asking for here? It seems like there are a few valid ways this could be done but it makes sense to me to email the old address. 

We should send an email to the address that is no longer primary.

lib/hexpm/accounts/audit_log.ex
```diff
@@ -248,6 +248,12 @@ defmodule Hexpm.Accounts.AuditLog do
   defp extract_params("password.reset.finish", nil), do: %{}
   defp extract_params("password.update", nil), do: %{}
 
+  defp extract_params("security_notification.email_added", params), do: params
+  defp extract_params("security_notification.primary_email_changed", params), do: params
+  defp extract_params("security_notification.api_key_created", params), do: params
+  defp extract_params("security_notification.api_key_revoked", params), do: params
+  defp extract_params("security_notification.api_keys_all_revoked", params), do: params
```
ericmj: All of these already have audit logs from the primary action that is causing the notification. There is no reason to add additional audit logs for this.

Ch4s3:
> > did I misunderstand what you were asking for here? It seems like there are a few valid ways this could be done but it makes sense to me to email the old address.
> 
> We should send an email to the address that is no longer primary.

Ok, that's what I thought.

lib/hexpm/accounts/audit_log.ex
Ch4s3: No problem. I wasn't sure how granular audit logs are supposed to be, but I'll keep this in mind in the future.

Ch4s3: done

lib/hexpm/emails/emails.ex
```diff
@@ -82,6 +82,51 @@ defmodule Hexpm.Emails do
...
+  def email_added(user, new_email) do
+    email()
+    |> email_to(user)
+    |> subject("Hex.pm - A new email address was added to your account")
+    |> assign(:username, user.username)
+    |> assign(:new_email, new_email.email)
+    |> render(:email_added)
+  end
+
+  def primary_email_changed(user, old_addr, new_addr) do
```
ericmj, line 94: We should only send to the old address.

Ch4s3: Sorry, this is a reading comprehension issue on my end I think.

Ch4s3: done

lib/hexpm/accounts/security_notifications.ex
```diff
@@ -0,0 +1,53 @@
+defmodule Hexpm.Accounts.SecurityNotifications do
```
ericmj: I feel like this module is a bit inconsistent with the rest of the application. Why are only some security notifications in here but not others? Generally we just inline email sending.

ericmj (approved):
Looks good, just one small comment.

lib/hexpm/accounts/security_notifications.ex
Ch4s3: I meant for this to be a proposal for moving all security notification into this module rather than having them mixed into other places. I should have been more clear in the PR description. The idea would be to move other instances over in a follow on if you like the idea, or fold these back into the other modules if you think its unnecessary.

ericmj: I would suggest folding these back into the other modules. 👍

ericmj:
Thank you! 💜

Issue #1581, opened 2026-05-19
Update hexpm Dockerfile to run newer versions of Elixir and Erlang.


Ch4s3:
I've been running the app on my machine on the latest OTP and Elixir without issue, so it looks like it might be safe and beneficial to start updating hex.pm. I'd suggest starting with a jump to Elixir 1.19.5 and later moving to OTP 28 once everyone is confident that things are working well.

Ch4s3:
#1583 closes this out

PR #1583, opened 2026-05-19
bump OTP to 28.5 and debian to trixie-20260505-slim

Ch4s3:
This PR bumps OTP to 28.5 per my issue #1581, and updates Debian to the next release of trixie because there wasn't a hexpm/elixir image with the old Debian version and the new versions of Elixir/Erlang. I'd have preferred to do it separately, but it shouldn't be an issue.

I also had to introduce language versions to the test matrix since older hex versions pin `cowlib 2.8.0` which isn't compatible with `OTP > 27`.


ericmj:
Thank you! 💜

PR #1594, opened 2026-05-23
Add Sobelow security scanning to CI [WIP]

Ch4s3:
## What

Adds the [Sobelow](https://github.com/nccgroup/sobelow) static analysis tool and runs it in CI, failing the build on any **new** security finding.

## Changes

- **`mix.exs` / `mix.lock`** — add `{:sobelow, "~> 0.14", only: [:dev, :test], runtime: false}` (0.14.1).
- **`.github/workflows/main.yml`** — add a `Security scan (Sobelow)` step to the `format` job (already compiles the app, so no extra build).
- **`.sobelow-conf`** — `exit: :low` (fail on any new finding), `skip: true` (honor the baseline), `private: true` (skip the network update-check in CI).
- **`.sobelow-skips`** — fingerprint baseline of the 76 pre-existing findings, generated via `mix sobelow --mark-skip-all`.

## baseline

The codebase has 76 existing findings (9 High / 7 Medium / 60 Low). Rather than ignore whole rule categories or files (which creates blind spots), the baseline records each finding by fingerprint (rule + file:line + hash). A bare `mix sobelow` passes today, but a *new* finding — even of an already-baselined rule, or the same finding on a moved line — still breaks the build.

The pre-existing findings are suppressed so the gate can land, **not** dismissed as safe. A follow-up to triage the High-confidence ones is worthwhile. (One already reviewed: `lib/hexpm/repo.ex:154` is a false positive — static SQL string with a `$1` placeholder and parameterized value.)

## Verification (run locally)

- `mix sobelow` (config-driven) → exit 0 with the baseline in place.
- Temporary canary (`String.to_atom/1` on input) → exit 1, confirming new issues break the build. Canary removed.
- `mix format --check-formatted` → clean.


Ch4s3:
This isn't ready for review

PR #1597, opened 2026-05-26
Show version requirement on dependants page

maennchen:
Maintainers can now see which version requirement each dependent package uses, making it easy to assess ecosystem adoption of latest releases and to coordinate CVEs across affected dependants.

Screenshot 2026-05-26 at 18 52 47
Screenshot 2026-05-26 at 18 52 58

(The seed postgrex entry actually requires `0.1.0`, this change also shows more realistic requirements like `~> X.Y`)


ericmj:
Thank you! 💜

PR #1603, opened 2026-05-27
Fix absolute image paths in READMEs not resolving to preview URL

maennchen:
Paths like `/images/foo.png` were returned unchanged because `Path.safe_relative/1` rejects leading slashes. Strip the leading `/` alongside `./` so they resolve against the package preview base URL.

This resolves the first part of #1602

### DIFF
--- a/lib/hexpm_web/readme/url_rewriter.ex
+++ b/lib/hexpm_web/readme/url_rewriter.ex
@@ -86,7 +86,10 @@ defmodule HexpmWeb.Readme.URLRewriter do
         url
 
       true ->
-        path = String.trim_leading(url, "./")
+        path =
+          url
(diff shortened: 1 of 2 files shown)


RaymondLoranger:
I removed the leading slash but it still does not work.

maennchen:
@RaymondLoranger It looks like the images are not added to the hex package:
https://github.com/RaymondLoranger/io_ansi_plus/blob/b824a8adf135d9ff8ed469a987f506f760ba04c4/mix.exs#L31

RaymondLoranger:
Many thanks Jonatan!

maennchen:
@RaymondLoranger Starting to look prettier 😄  We'll get the rest sorted out as well.

PR #1607, opened 2026-05-28
Disable Earmark HTML escaping in favor of downstream sanitization

### LINKED ISSUE #1602: Footnote and image don't work in README.md
In a README.md, the footnote notation [^1] and [^1]: "Footnote content" does not work in hex.pm.
See https://hex.pm/packages/io_ansi_plus under section `Notes`.

Also image links like the following ones do not work:

```
![!](/images/00ffff.png) `:aqua`
![examples](/images/examples.png)
```

However, the _same_ README.md file is correctly rendered in GitHub.
See https://github.com/RaymondLoranger/io_ansi_plus

maennchen:
Earmark's transform/2 escapes inline HTML tags like `<ins>` within paragraphs by default, causing them to render as literal text rather than HTML. Since the sanitizer already strips disallowed tags downstream, Earmark's escaping is redundant and can be safely disabled with `escape: false`.

Resolves https://github.com/hexpm/hexpm/issues/1602#issuecomment-4559211439

### DIFF
--- a/lib/hexpm_web/controllers/readme_controller.ex
+++ b/lib/hexpm_web/controllers/readme_controller.ex
@@ -109,7 +109,7 @@ defmodule HexpmWeb.ReadmeController do
       case ext do
         ext when ext in [".md", ".markdown"] ->
           {_status, ast, _messages} = Earmark.Parser.as_ast(content, gfm: true)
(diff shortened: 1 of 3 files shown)


maennchen:
Superseded by #1608

PR #1638, opened 2026-06-06
Show country geography of login IP on the audit log page

Ch4s3:
## Summary

- Adds `Hexpm.Geo` behaviour + `Hexpm.Geo.Geolix` production implementation (via `geolix` + `geolix_adapter_mmdb2`) that resolves an IP to a `%{iso_code, name}` country map
- Renders a flag emoji + country name next to each audit-log timestamp on `/dashboard/audit-logs` and the organisation audit-log tab
- Uses the free [DB-IP IP-to-Country Lite](https://db-ip.com/db/download/ip-to-country-lite) database (CC BY 4.0; attribution rendered on the page); any MaxMind-compatible MMDB country database works equivalently

## Zero-config deployment

The Docker image bakes the database in at build time (`mix download_geoip`). The app auto-discovers it via `:code.priv_dir/1` at release startup — nothing extra to configure. Non-Docker deploys can point at their own file via `HEXPM_GEOIP_COUNTRY_PATH`. If no database is found, geolix logs an info message and the app boots normally with locations omitted.

## Implementation notes

- `Hexpm.Geo.Geolix.parse_result/1` handles both result shapes geolix can return: a `Result.Country` struct (recognised MaxMind `database_type`) and the raw decoded map that DB-IP produces (unrecognised `database_type`). The `names → :en` fallback is the only source of a human-readable name on the DB-IP path — not dead code.
- Dev/test use `Hexpm.Geo.Local` (no database required); localhost is mapped by default. `Hexpm.Geo.Mock` (Mox) is wired into `ConnCase` so no per-file setup is needed in existing tests.
- `mix download_geoip` fetches the current month's DB-IP file and writes it to `priv/geoip/country.mmdb`. Pass `--month YYYY-MM` if this month's file isn't published yet.


ericmj:
Can you include screenshots of what this looks like in desktop, mobile, and dark mode? Thank you!

Ch4s3:
@ericmj I'll add those in a bit. I meant to open this as a draft, I'm still messing around with it.

Ch4s3:
Screenshot 2026-06-06 at 5 02 12 PM
Screenshot 2026-06-06 at 5 02 25 PM
Screenshot 2026-06-06 at 5 02 48 PM
Screenshot 2026-06-06 at 5 03 01 PM

ericmj:
Thanks for the screenshots. It would be great to document (maybe in README or somewhere else) how to source the geo database and set it up.

Let us know when it's ready for review.

Ch4s3:
@ericmj documenting setup is definitely on my todo list. It still feels a little speculative, but if everyone likes this, it might be nice to add login locations to the sign in emails.

Ch4s3:
@ericmj I think this is ready for review. The Docker change is worth some consideration. I'm not sure if you'd prefer to build it this way, or have something you can run against the console.

priv/repo/seed_demo_audit_logs.exs
```diff
@@ -0,0 +1,48 @@
+# One-shot seeder for demo audit-log entries used to capture screenshots
```
ericmj: What's the use case for this case file? If we need some seed data for local testing you can put it in `seeds.exs`.

Dockerfile
```diff
@@ -36,6 +36,11 @@ COPY lib lib
 RUN mix assets.deploy
 RUN mix compile
 
+# bundle the IP geolocation database into the release (priv/geoip/country.mmdb).
+# Optional and best-effort: if the download fails the build still succeeds and
+# geolocation is simply disabled at runtime (see config/runtime.exs).
+RUN mix download_geoip || echo "geoip download failed; IP geolocation will be disabled"
```
ericmj: Falling back seems fragile. If there is a network error during docker build now my image suddenly won't have geodata.

lib/mix/tasks/download_geoip.ex
```diff
@@ -0,0 +1,68 @@
...
+
+  @default_output "priv/geoip/country.mmdb"
+  @base_url "https://download.db-ip.com/free"
+
+  @impl Mix.Task
+  def run(args) do
+    {opts, _, _} = OptionParser.parse(args, strict: [output: :string, month: :string])
+    output = opts[:output] || @default_output
+    month = opts[:month] || Calendar.strftime(Date.utc_today(), "%Y-%m")
+    url = "#{@base_url}/dbip-country-lite-#{month}.mmdb.gz"
```
ericmj: Will this always be available at the first day of the month? What's their release cadence?

Dockerfile
ericmj: There is nothing to bust the docker cache for this. Maybe we should pass in an explicit month here? Maybe a build arg that gets passed in for the current month?

lib/hexpm_web/templates/dashboard/audit_log/index.html.heex
```diff
@@ -24,6 +24,8 @@
             )}
           </div>
         <% end %>
+
+        <.dbip_attribution />
```
ericmj, line 28: We need to add this everywhere we show audit logs, like the organization dashboard.

lib/hexpm/geo/geolix.ex
```diff
@@ -0,0 +1,63 @@
...
+      `names[locale]`.
+    * Unrecognized (e.g. DB-IP, whose `database_type` is not in geolix's
+      mapping) — the raw decoded map, where `:name` is NEVER set and only
+      `:names` is present.
+
+  The DB-IP path is therefore always the raw-map path, so the `names` →
+  English fallback in `country_name/2` is REQUIRED, not dead code: it is the
+  only source of a human name there. `fetch/2` additionally accepts atom OR
+  string keys, so the parser stays correct regardless of the decoder's
+  `map_keys` option or a `result_as: :raw` lookup.
```
ericmj: What is `result_raw: :raw`? I don't see it in the code anywhere.

lib/hexpm/geo/geo.ex
```diff
@@ -0,0 +1,34 @@
...
+  def lookup_country(nil), do: nil
+  def lookup_country(ip) when is_binary(ip), do: impl().lookup_country(ip)
+
+  @doc """
+  Converts a two-letter ISO 3166-1 alpha-2 country code into its flag emoji
+  using Unicode regional indicator symbols. Returns "" for any other input.
+  """
+  @spec flag_emoji(String.t()) :: String.t()
+  def flag_emoji(<<a, b>>) when a in ?A..?Z and b in ?A..?Z do
+    <<127_397 + a::utf8, 127_397 + b::utf8>>
```
ericmj, line 30: This is neat :)

Ch4s3: yeah, its fun I love a binary pattern match.

priv/repo/seed_demo_audit_logs.exs
Ch4s3: will do

Dockerfile
Ch4s3: Ok,I'll think about this.

lib/mix/tasks/download_geoip.ex
Ch4s3: Monthly on the 1st https://db-ip.com/faq.php

priv/repo/seed_demo_audit_logs.exs
Ch4s3: Moved into `seeds.exs` — the entries are appended inside the existing transaction, reusing the `eric` user that's already seeded there. Separate file deleted.

Dockerfile
Ch4s3: Removed the fallback — network errors now fail the build. Addressed the cache-busting point in the other comment.

Ch4s3: Added `ARG GEOIP_MONTH` to the build stage. Passing `--build-arg GEOIP_MONTH=YYYY-MM` to `docker build` fetches a specific release and busts the layer cache. If omitted, the task defaults to the current month (same behaviour as before).

lib/mix/tasks/download_geoip.ex
Ch4s3: DB-IP publishes a new file on the 1st of each month, so the current-month file is available from day one. The `--month` flag (and now the Docker build arg) is there as an escape hatch if needed, e.g. if a build runs before the new file is published, or to pin a specific release.

lib/hexpm_web/templates/dashboard/audit_log/index.html.heex
Ch4s3: Added — `dbip_attribution` is now imported and called inside `audit_logs_tab/1`, so it renders on the organisation dashboard tab too.

lib/hexpm/geo/geolix.ex
Ch4s3: Removed the stale reference — that option was refactored out and the doc wasn't updated. Fixed.

lib/hexpm/geo/geo.ex
maennchen: Unfortunately not quite right. The DB contains both `ZZ` (unknown) and `XK` (Kosovo).

Kosovo depends on the OS. ZZ will print `🇿🇿 `.

maennchen: Checked with:

```bash
cat dbip-country-lite-2026-06.csv.gz | gunzip | awk -F, '{print $3}' | sort -u | while read -r code; do
  a=$(printf '%d' "'${code:0:1}")
  b=$(printf '%d' "'${code:1:1}")
  printf '%s - %b%b\n' \
    "$code" \
    "\\U$(printf '%08X' $((127397 + a)))" \
    "\\U$(printf '%08X' $((127397 + b)))"
done
```

Ch4s3: on it

Ch4s3: Fixed — `parse_result/1` now treats `ZZ` the same as a missing country (returns `nil`), so unknown IPs show no location. Added a regression test.

`XK` (Kosovo) is left as-is: it's a real country and renders the Kosovo flag on macOS, iOS, and Android; on systems that don't support the regional indicator pair it falls back to rendering the letters, which is still informative.

maennchen: > Emoji flags are supported on all major platforms except [Windows](https://emojipedia.org/microsoft/), which displays two-letter country codes instead of emoji flag images. Windows has never supported any national flag emoji design.

So Windows (and sometimes Kosovo) will display letters.

Do you have access to a windows device so that we can see the result? (I have none to test)

PS: I didn’t want to express an opinion about Kosovo not being a country. But because of the dispute it might not render in some places.

Ch4s3: I can check on a windows machine via ngrok I think. I would also like to refrain from commenting on Kosovo which I know very little about.

Ch4s3: On the left is firefox and chrome on the right. So the flags display on FireFox on windows but not Chrome.
Image

Edge refused to connect over the Cloudflare tunnel I used, but its basically chromium so probably doesn't display the flags. I could detect the OS with a line of JS and conditionally hide the flag but that seems overkill if there aren't a lot of windows users. Based on [this Gleam survey](https://gleam.run/news/developer-survey-2024-results/) there are at least some windows users.

Ch4s3:
The CI failure in `Test Hexpm` is in `release_controller_test.exs`, not related to this PR. The test uses `Fake.sequence(:package)` to generate a package name; the upstream `1d39da31` commit added org-name reservation logic, and when the sequence counter lands on a reserved name the test gets a 422 instead of 201. Main passes because its counter is at a different state. Nothing in this PR touches package validation or the release controller.

lib/mix/tasks/download_geoip.ex
ericmj: They say "Database downloads are updated monthly usually on the 1st of each month." so we need to fall back to last month to be safe.

ericmj:
Do you have a screenshot on what the attribution looks like?

Ch4s3:
> Do you have a screenshot on what the attribution looks like?

image


I think this is both sufficient and not to obtrusive.

ericmj:
@Ch4s3 let us know when this is ready for another review. I think there is still some work ot make adding the geo database to the docker image less fragile.

We want the docker build to fail if we cannot add the database and we need to fall back if the database for the current month hasn't been added yet.

lib/mix/tasks/download_geoip.ex
```diff
@@ -0,0 +1,89 @@
...
+        download(prev, output, true)
+
+      {:ok, %{status: 404}} ->
+        Mix.raise("Not found (404): #{url}")
+
+      {:ok, %{status: status}} ->
+        Mix.raise("Download failed (HTTP #{status}): #{url}")
+
+      {:error, reason} ->
+        Mix.raise("Download failed: #{inspect(reason)}")
```
Ch4s3, line 61: @ericmj I think this addresses your request hard fails on when the data can't be found

lib/mix/tasks/download_geoip.ex
```diff
@@ -0,0 +1,89 @@
...
+        write(body, output)
+
+      {:ok, %{status: 404}} when not explicit_month? ->
+        prev = previous_month(month)
+
+        Mix.shell().info(
+          "#{month} not published yet — falling back to #{prev}."
+        )
+
+        download(prev, output, true)
```
Ch4s3, line 52: this will look back once

Ch4s3:
@ericmj I think it's good for review now. Let me know if the fallback strategy doesn't look right to you.

Ch4s3:
@ericmj did you need other changes on this (other than conflicts)?

ericmj:
@Ch4s3 I've been trying to decide if we should go ahead with this. I am afraid it's adding a lot of code for the scope of the feature, but I think it's worth it for security.

I will go ahead and merge this. Do you think we should add it to the "Last Activity" panels for Keys and Sessions in the dashboard also?

Ch4s3:
@ericmj I'd probably go with both locations.

If it proves to be a hassle to maintain, I'd take no offense if its removed. I'm also happy to do maintenance work personally.

Issue #1590, opened 2026-05-20
Potentially adding Sobelow to CI


Ch4s3:
Has there ever been any discussion of adding [Sobelow](https://github.com/nccgroup/sobelow) as a CI check? I've used it on other projects as a quick sanity check for things like SQL injection, directory traversal, XSS, and so on. I haven't found false positives to be too bad in general, and it occasionally catches PRs that have subtle issues.

ericmj:
I don't think there's been a discussion and I haven't used it myself so I can't really say if it's something that would be useful for us.

Ch4s3:
I can put together a draft PR later and see if it turns anything up.

PR #1640, opened 2026-06-09
Add Sobelow security scanning to CI

Ch4s3:
Adds [Sobelow](https://github.com/nccgroup/sobelow) as a CI check, as floated in #1590. Sobelow is NCC Group's static analyzer for Phoenix — it flags SQL injection, directory traversal, XSS, missing CSP, hardcoded secrets, and unsafe atom creation by walking the source AST.

Run against the current tree, it surfaces ~70 findings but **nothing actionable**: all 9 high-confidence results are false positives (detailed below), and the rest are low/medium noise. So this isn't catching existing bugs — the value, if we want it, is a regression guard on the request-handling code, at the cost of maintaining suppressions. Putting it up as a draft so we can judge that trade-off against real output rather than in the abstract.

Wiring:
- `{:sobelow, "~> 0.14", only: [:dev, :test], runtime: false}`.
- A "Security analysis (sobelow)" step in the `format` job runs `mix sobelow --skip --exit Low`, which fails only on findings not already baselined — i.e. new code.
- Existing findings are suppressed in two layers. The 9 high-confidence ones are documented: inline `# sobelow_skip` comments on the two function-scoped findings (`test_controller.ex`, `docs_controller.ex`), and a commented `.sobelow-conf` for the config/router findings, which sobelow won't accept inline annotations for. The remaining 7 medium + 61 low are fingerprinted in `.sobelow-skips` for later triage.

The high-confidence false positives:
- `Config.Secrets` ×3 in `config/hex.exs` — test config; the "secrets" are fixtures (`postgres`, a hex-encoded `youfoundthehex`). Prod reads from env.
- `Config.CSP` ×3 on the `browser`/`browser_api`/`readme` pipelines — CSP is set by our own plugs (`ContentSecurityPolicy`, `ReadmeContentSecurityPolicy`), which sobelow can't see; it only recognizes a literal map passed to `put_secure_browser_headers`.
- `Config.HTTPS` in `config/prod.exs` — TLS terminates at the edge, so `force_ssl` is intentionally absent (`url: [scheme: "https"]`).
- `XSS.SendResp` in `test_controller.ex` — only routed under `Mix.env() in [:dev, :test, :hex]`, never prod.
- `Traversal.FileModule` in `docs_controller.ex` — the path is the Plug-written temp file for the upload body, not a user-supplied string.

One wart: `ignore: ["Config.CSP"]` is global, so sobelow won't re-flag CSP if a pipeline's plug is later removed. Given the plugs are structural to the router I think that's acceptable, but it's a real gap.


Ch4s3:
proposal for #1590

.sobelow-skips
```diff
@@ -0,0 +1,91 @@
...
+XSS.Raw: XSS,lib/hexpm_web/templates/readme/show.html.heex:13,57C6BB9
+XSS.Raw: XSS,lib/hexpm_web/templates/email/api_key_revoked.html.eex:14,595BB7D
+XSS.Raw: XSS,lib/hexpm_web/views/blog_view.ex:28,6A78B8D
+SQL.Query: SQL injection,lib/hexpm/repo.ex:176,6BC2C68
+XSS.Raw: XSS,lib/hexpm_web/markdown_engine.ex:30,700DF56
+DOS.StringToAtom: Unsafe `String.to_atom`,lib/hexpm/accounts/users.ex:527,7CEF88F
+RCE.CodeModule: Code Execution in `Code.eval_file`,lib/hexpm/release_tasks.ex:250,7E79308
+XSS.Raw: XSS,lib/hexpm/repository/packages.ex:446,9D87A4
+XSS.Raw: XSS,lib/hexpm_web/templates/email/api_keys_all_revoked.html.eex:14,ABD0B8
+DOS.StringToAtom: Unsafe `String.to_atom`,lib/hexpm/accounts/users.ex:505,EBC452
```
ericmj, line 91: I'm not a fan of the external `file:line` annotations, code modifications in these files means we constantly have to update this file. Can we make them inline like above for `# sobelow_skip ["Traversal.FileModule"]`?

ericmj (commented):
If all the findings are false positives I guess there is a question how much value this adds. But maybe that's unfair just after we had a security audit?

ericmj:
@maennchen What are your thoughts on this?

maennchen:
I have not gone through all ignores, but it they are all false positives and sobelow found 0 actual issues, The noise ratio is way too high for this to be practical.

Ch4s3:
Yeah, I was hoping the output would be less noisy. However to clarify these skips are not all false positives, they're existing _potential_ issues that need to be checked individually. The PR is a draft for this reason. I need to go through and check them all. The tool can for example see that a raw string going into an Ecto query can only ever come from a compile time string. 

There's probably a good amount of work here in manually running down all of the warnings and finding the right threshold for alerting. I've used this tool successfully before, but the on boarding for any larger existing app is a pain.  I'm open to working on it, but only if you both think there's potential value here.

ericmj:
If you could review some of those findings and check if there are any true positives, that would be helpful. If it turns the vast majority are false positives then I don't think we want to merge this.

Thanks for investigating!

ericmj:
@Ch4s3 what is the status of this PR?

Ch4s3:
@ericmj I think the load of false positives on old code was/is too high to get the value out of sobelow here. I'm closing this for now.

ericmj:
Thanks for looking into it.

PR #1647, opened 2026-06-11
Fix intermittent failure in release controller test

Ch4s3:
## Summary

- `Fake.sequence(:package)` draws names from `priv/fake/packages.txt`. Commit `0be1dd4f` added `stats` to `@subdomain_names`, making it a reserved package name — but it was still present in `packages.txt`.
- When the sequence counter landed on `stats` the request returned HTTP 422 `{"name":"is reserved"}`, causing a flaky test failure.
- Fix: remove `stats` from `priv/fake/packages.txt`. It is the only entry that overlaps with the full `@reserved_names` list.

### DIFF
--- a/priv/fake/packages.txt
+++ b/priv/fake/packages.txt
@@ -3262,7 +3262,6 @@ stationary
 statistics
 statix
 statman
-stats
 statsderl
 status_ku
 std_json_io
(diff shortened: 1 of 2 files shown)


ericmj:
We should remove the reserved package names from `priv/fake/packages.txt`.

PR #1653, opened 2026-06-12
Allow users to delete their account

ericmj:
Adds self-service account deletion under /dashboard/delete-account.

Deleting an account is a hard delete of the users row. Audit logs and packages are preserved through the existing FK rules: releases keep existing with a NULL publisher, audit logs keep their user_data snapshot. Keys, emails, sessions and org memberships are cascade deleted. The username goes into a new reserved_usernames table and can never be registered again, by a user or an organization, so nobody can impersonate the old account or appear to inherit its packages.

Users who are the last admin or only member of an organization cannot delete their account until they hand the organization over. Being the sole owner of packages only warns: those packages become ownerless but stay installable, and new owners can only be assigned by us.

The flow requires a fresh sudo re-authentication (a new force option on the sudo plug that ignores the rolling window granted at login), typing your username, and then clicking a single-use link emailed to the primary address, valid for 24 hours. The link only works from a logged in session of the same user, so a leaked or stolen key does nothing on its own. Changing your password or primary email cancels a pending request, and request creation is throttled and audit logged.

AdminTasks.remove_user now goes through the same code path, so admin removals also reserve the username and write an audit log. The request table is cleaned up by the daily purge task.

### DIFF
--- a/lib/hexpm/release_tasks/purge_expired_records.ex
+++ b/lib/hexpm/release_tasks/purge_expired_records.ex
@@ -122,7 +122,7 @@ defmodule Hexpm.ReleaseTasks.PurgeExpiredRecords do
         repo,
         Hexpm.Accounts.AccountDeletionRequest,
         from(r in Hexpm.Accounts.AccountDeletionRequest,
-          where: r.inserted_at < fragment("NOW() - make_interval(days => ?)", @retention_days)
+          where: r.inserted_at < ago(@retention_days, "day")
         ),
         batch_size
       )
--- a/lib/hexpm/release_tasks/purge_expired_records.ex
+++ b/lib/hexpm/release_tasks/purge_expired_records.ex
@@ -23,6 +23,7 @@ defmodule Hexpm.ReleaseTasks.PurgeExpiredRecords do
       purge_user_sessions(repo, batch_size)
       purge_plug_sessions(repo, batch_size)
       purge_password_resets(repo, batch_size)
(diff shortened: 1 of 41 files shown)


lib/hexpm/release_tasks/purge_expired_records.ex
```diff
@@ -115,6 +116,20 @@ defmodule Hexpm.ReleaseTasks.PurgeExpiredRecords do
     Logger.info("[task] Purged #{count} expired password resets")
   end
 
+  defp purge_account_deletion_requests(repo, batch_size) do
+    count =
+      delete_in_batches(
+        repo,
+        Hexpm.Accounts.AccountDeletionRequest,
+        from(r in Hexpm.Accounts.AccountDeletionRequest,
+          where: r.inserted_at < fragment("NOW() - make_interval(days => ?)", @retention_days)
```
maennchen:
```suggestion
          where: r.inserted_at < ago(@retention_days, "day")
```

maennchen (commented):
`priv/repo/migrations/20230124011453_add_key_id_to_audit_logs.exs:6`

Audit Log / Key ID is `ON DELETE RESTRICT` => May prevent users from deleting their account.

ericmj:
We fixed the constraint here: https://github.com/hexpm/hexpm/blob/main/priv/repo/migrations/20260206120000_fix_audit_logs_key_delete_constraint.exs#L5

PR #1666, opened 2026-06-19
Add API endpoint to retire all active package releases

### LINKED ISSUE #1650: Retire Complete Package
`mix hex.retire` allows to retire releases, but not to retire a whole package.

If for example maintenance on a package is given up, it would be nice to have an API / CLI Command to retire all versions at the same time.

Unclear is how you would deal with already retired versions. Should they be overwritten with the new retirement or just left alone?

BittuBarnwal7479:
Implements the server-side API portion of #1650 by adding support for retiring all active releases of a package in a single request.

Resolved: #1650

### DIFF
--- a/lib/hexpm/repository/releases.ex
+++ b/lib/hexpm/repository/releases.ex
@@ -138,11 +138,19 @@ defmodule Hexpm.Repository.Releases do
       Enum.reduce(releases, Multi.new(), fn release, multi ->
         key = {:release, release.id}
 
-        multi
-        |> Multi.update(key, Release.retire(release, params))
-        |> audit(audit_data, "release.retire", fn changes ->
-          {package, Map.fetch!(changes, key)}
-        end)
+        Multi.update(multi, key, Release.retire(release, params))
+      end)
+      |> Multi.merge(fn changes ->
+        retirements =
+          Enum.map(releases, fn release ->
+            {package, Map.fetch!(changes, {:release, release.id})}
+          end)
+
+        if retirements == [] do
+          Multi.new()
+        else
+          audit_many(Multi.new(), audit_data, "release.retire", retirements)
+        end
       end)
     end)
     |> Repo.transaction(timeout: @publish_timeout)
--- a/lib/hexpm/repository/releases.ex
+++ b/lib/hexpm/repository/releases.ex
@@ -119,7 +119,9 @@ defmodule Hexpm.Repository.Releases do
     |> retire_result()
   end
 
-  def retire(package, params, audit: audit_data) do
+  def retire(package, params, opts) do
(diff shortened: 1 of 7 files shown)


ericmj:
I think the current default that keeps existing retirements is good but do we want an option to also overwrite all existing retirements?

BittuBarnwal7479:
> I think the current default that keeps existing retirements is good but do we want an option to also overwrite all existing retirements?

good point, implemented optional overwrite behavior.

Default behavior still preserves existing retirements. Added `replace: true` to overwrite retirement metadata across all releases.

`replace` is an operation option only and is not persisted. Overwrites generate audit entries and trigger a registry rebuild.

Added tests for both modes. Test suite passes (33 tests, 0 failures).

lib/hexpm/repository/releases.ex
```diff
@@ -119,6 +119,36 @@ defmodule Hexpm.Repository.Releases do
...
+
+      {:ok, repo.all(query)}
+    end)
+    |> Multi.merge(fn %{releases: releases} ->
+      Enum.reduce(releases, Multi.new(), fn release, multi ->
+        key = {:release, release.id}
+
+        multi
+        |> Multi.update(key, Release.retire(release, params))
+        |> audit(audit_data, "release.retire", fn changes ->
```
ericmj: Let's use audit_many instead to avoid extra INSERTs

lib/hexpm_web/controllers/api/retirement_controller.ex
```diff
@@ -1,10 +1,32 @@
 defmodule HexpmWeb.API.RetirementController do
   use HexpmWeb, :controller
 
+  plug :maybe_fetch_package when action in [:create_package]
   plug :maybe_fetch_release when action in [:create, :delete]
 
   plug :authorize, domains: [{"api", "write"}, "package"], fun: {AuthHelpers, :package_owner}
 
+  def create_package(conn, params) do
```
ericmj: Maybe this name is a bit more clear?

```suggestion
  def create_all(conn, params) do
```

ericmj (approved):
Two minor comments, otherwise looks good.

lib/hexpm/repository/releases.ex
BittuBarnwal7479: > Let's use audit_many instead to avoid extra INSERTs

thanks for pointing, updated to use `audit_many` for the package-level retire audit logs.

lib/hexpm_web/controllers/api/retirement_controller.ex
BittuBarnwal7479: Sure :-)

PR #1677, opened 2026-06-30
disable autorecover on the filter and sort forms

### LINKED ISSUE #1676: Browsing all packages periodically resets the view
When you are browsing all packages on the site (https://hex.pm/packages?sort=recent_downloads) and are on a different page than page 1 it resets the view and takes you back to page 1. This happens when you stay at the page for a little while and my best clue to what is happening is that can see that the websocket resets periodically. (See screenshot) My current guess to why it resets page is the url state with the page isn't handled correctly and for the websocket reset I don't know why they happens.

Image

MikaelFangel:
Fixes #1676 

When the connection is reset and because the filter has both `phx-change` and an `id` it auto recovers the form (docs: https://phoenix-live-view.hexdocs.pm/form-bindings.html#recovery-following-crashes-or-disconnects) which then results in a change event that resets the page because of the handler. I also added the same for sort even though the bug is not currently present I thought that would be the defensive choice.

To test in the chrome browser before and after the change you can write this in the console from a page that is not the first: `liveSocket.disconnect(); liveSocket.connect()`

### DIFF
--- a/lib/hexpm_web/live/package_live/filter_sidebar.ex
+++ b/lib/hexpm_web/live/package_live/filter_sidebar.ex
@@ -132,7 +132,8 @@ defmodule HexpmWeb.PackageLive.FilterSidebar do
 
   defp filter_form(assigns) do
     ~H"""
-    <form phx-change="filter_change" id={@id}>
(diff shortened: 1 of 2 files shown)


ericmj:
Thank you! 💜

PR #1695, opened 2026-07-09
Restore form recovery on package index

rhcarvalho:
And prevent page reset on reconnect.

This is another take on #1676, as I casually noticed that by disabling form recovery we ended up breaking other common use cases, such as if a user types or updates filters while (potentially briefly) offline the browser-local inputs are discarded.

Instead of ignoring form recovery, we check if the incoming event payload represents an actual change relative to the current assigns. If they are identical, we return `{:noreply, socket}` without patching the URL. This preserves the page state on reconnect when no local form changes happened, while still allowing explicit user changes to correctly reset the page.

It also reduces the amount of patches in the normal connected case, since we don't patch when nothing is supposed to change (if load is actually a problem we can also add `phx-debounce` to reduce the frequency of events as users type).

Additionally, LiveView form recovery requires forms with `phx-change` to have an `id`, so that was added where missing.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -24,6 +24,7 @@ import { PrivateRepoTabs } from "./hooks/private_repo_tabs";
 import { PolicyDirtyState } from "./hooks/policy_dirty_state";
 import LineHighlight from "./hooks/line_highlight";
 import { InfiniteScroll } from "./hooks/diff";
+import { FormSync } from "./hooks/form_sync";
 
 let csrfToken = document
   .querySelector("meta[name='csrf-token']")
@@ -51,6 +52,7 @@ let Hooks = {
   PolicyDirtyState,
   LineHighlight,
   InfiniteScroll,
+  FormSync,
 };
 let liveSocket = new LiveSocket("/live", Socket, {
   params: { _csrf_token: csrfToken },
--- a/assets/js/hooks/form_sync.js
+++ b/assets/js/hooks/form_sync.js
@@ -1,0 +1,41 @@
+/**
+ * FormSync Hook
+ *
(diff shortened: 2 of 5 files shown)


rhcarvalho:
Rebased on top of latest main, applied missing `mix format` in tests.

ericmj (commented):
Both the mobile and desktop forms are still in the DOM, so LiveView will recover both of them. This can undo the value we were trying to recover: for example, the mobile sort form replays name, then the stale desktop form replays recent_downloads and patches it straight back. The same thing happens with desktop filters followed by the stale mobile filter form.

Could we give sort and filter their own phx-auto-recover events and ignore the second changed recovery payload for each group? Both form orders reproduce locally; the normal same-value guard is not enough once the first recovery has updated the assigns.

rhcarvalho:
Hi @ericmj, thanks for the review. Good catch, you are right that form recovery would race. Sorry for the delay, I was mostly offline last week.

> Could we give sort and filter their own phx-auto-recover events and ignore the second changed recovery payload for each group? Both form orders reproduce locally; the normal same-value guard is not enough once the first recovery has updated the assigns.

Thanks for the suggestion. I did that (https://github.com/praialabs/hexpm/commit/aa3011b795a1211af5e2e3a7d61d6243696cf4aa), and then realized that the race would remain: since the DOM forms are out of sync, there is no single source of truth, and thus picking the "correct" recovery event on the server side is not as simple as discarding anything but the first event.

Instead, I went with a similar approach used in the search query input which is kept in sync using a JS hook. It has the added benefit that the client-side state is kept in sync even while the socket is disconnected, and coming back online updates the URL/assigns to match the client state.

Rebased on top of latest main.

ericmj:
Thank you! 💜

PR #1697, opened 2026-07-09
Enable auto-recovery and pagination protection for package search

rhcarvalho:
Adds unique HTML IDs to mobile search forms so Phoenix LiveView can track and recover form state upon reconnection.

Additionally, compares the search parameter with the assigned search query in the `search_change` handler. This avoids redundant URL patches, preventing the user's pagination state from resetting to page 1 during auto-recovery or identical change events.

Similar to #1695, but because it is orthogonal to the original problem I'm sending as a separate PR. I can rebase later to merge the test cases into a single `describe` block.

### DIFF
--- a/lib/hexpm_web/components/navbar.ex
+++ b/lib/hexpm_web/components/navbar.ex
@@ -223,10 +223,14 @@ defmodule HexpmWeb.Components.Navbar do
     <div id="mobile-search-bar" class="hidden lg:hidden! bg-grey-800 pb-4">
       <div class="flex items-center gap-2">
         <form
+          id="mobile-search-bar-form"
           role="search"
           action={~p"/packages"}
           class="flex-1"
+          phx-change={if @live_search, do: "search_change"}
           phx-submit={@live_search && "search_submit"}
+          phx-hook={@live_search && "FormSync"}
+          data-sync-to={@live_search && "mobile-menu-search-form"}
         >
           <div class="relative">
             <div class="absolute left-3 top-1/2 -translate-y-1/2 pointer-events-none">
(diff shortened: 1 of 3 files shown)


ericmj:
@rhcarvalho Can you enable "Allow edits from maintainers" on this and your other PRs?

rhcarvalho:
Rebased on top of latest main, applied missing `mix format` in tests.

---


> @rhcarvalho Can you enable "Allow edits from maintainers" on this and your other PRs?

I looked into it, and it is not available. I believe it is because the fork is not in my personal account (https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork), so there's no checkbox in the usual place.

image

rhcarvalho:
Test failure is https://github.com/hexpm/hexpm/actions/runs/29228187597/job/86746628178?pr=1697

>   1) test purge_key/2 calls purge endpoint 3 times after waiting (Hexpm.CDN.FastlyTest)
> Error:      test/hexpm/cdn/fastly_test.exs:9
>      Assertion failed, no matching message after 1000ms
>      The process mailbox is empty.

Seems unrelated to this change, though it didn't fail in other recent PRs.

ericmj (commented):
Adding IDs is not enough here because LiveView only registers form[phx-change] for recovery. phx-change is still on the inputs, so neither mobile form will be replayed after reconnect.

Could we move phx-change to the forms and use a separate phx-auto-recover event? Since both mobile forms are present in the DOM, that handler also needs to ignore the stale second payload or it can undo the recovered term.

rhcarvalho:
> Adding IDs is not enough here because LiveView only registers form[phx-change] for recovery. phx-change is still on the inputs, so neither mobile form will be replayed after reconnect.

Oh yeah thank you, good catch, my bad.

Rebased on top of latest main, and updated to move `phx-change` to the form level, and use the new `FormSync` hook introduced in #1695 to keep the two forms that emit `search_change` in sync on the client side.

---

While reviewing this change I noticed an unexpected UX: both search forms can be made to render simultaneously, one next to the other. I'm addressing that in a separate follow up to keep the diff focused.

<details>

image


</details>

ericmj:
Thank you! 💜

PR #1700, opened 2026-07-09
Remove unused `sort_selector` component

rhcarvalho:
Removes the unused `sort_selector/1` component from `HexpmWeb.Components.Package` and deletes the module since it was the only remaining component.

Its usage was abandoned in commit 8b08ea16 (#1573), where sorting was instead inlined directly in the templates to support distinct mobile and desktop select forms.

If a similar unified component is introduced in the future, it should include a unique form ID attribute to participate correctly in LiveView form auto-recovery on reconnection (similar to #1697 and #1698).


ericmj:
Thank you! 💜

PR #1718, opened 2026-07-17
Load package diff gaps on demand

ericmj:
Package diff navigation can load individual files out of sequence, leaving unloaded ranges that the previous single end-of-list sentinel could not fill while scrolling upward.

This renders a lazy-loading sentinel for every contiguous unloaded range. Ranges before the selected file load backward and ranges after it load forward, with each request capped at five files. Loaded pieces remain in canonical order, and stale gap events are ignored.

Validated with the focused Diff LiveView tests, the full Hexpm suite (28 properties and 2,088 tests), formatting, `git diff --check`, and `mix assets.deploy`.

### DIFF
--- a/assets/js/hooks/diff.js
+++ b/assets/js/hooks/diff.js
@@ -5,7 +5,7 @@ export const InfiniteScroll = {
       ([entry]) => {
         if (entry.isIntersecting && !this.pending) {
           this.pending = true;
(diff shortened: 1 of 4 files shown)


ericmj:
Replaced by https://github.com/hexpm/hexpm/pull/1719.

PR #1720, opened 2026-07-17
Include Gleam revert+retire instructions in email

lpil:
Hello!

I figured since Gleam instructions are now included in other parts of the site it would be good to have instructions in this email too. I hope that's alright!

Thanks,
Louis


ericmj:
Thank you! 💜

PR #1737, opened 2026-07-20
Fix dual search field on mobile

rhcarvalho:
Clicking both the search icon and the hamburger menu icon on mobile would show two search fields on screen at once, which is a confusing UX.

This adds JS commands to only show one search field at a time. And, because the two fields also had slightly different styling, we also update the styling to match, reducing layout shifts when toggling between them.

## Before

image

## After (side-by-side for CSS comparison)

127 0 0 1_4000_packages_sort=updated_at search=nerves

## After (only one visible at a time, reduced layout shift)

127 0 0 1_4000_packages_sort=updated_at search=nerves (1)
127 0 0 1_4000_packages_sort=updated_at search=nerves (2)

---

_Related to https://github.com/hexpm/hexpm/pull/1697#issuecomment-5022056561_

### DIFF
--- a/lib/hexpm_web/components/navbar.ex
+++ b/lib/hexpm_web/components/navbar.ex
@@ -249,7 +249,7 @@ defmodule HexpmWeb.Components.Navbar do
           type="button"
           phx-click={show_modal("search-cheatsheet")}
           aria-label="Search filter cheatsheet"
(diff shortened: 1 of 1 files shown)


ericmj:
Nice find, thanks!

PR #1738, opened 2026-07-20
Cleanup component global attribute `:include`

rhcarvalho:
The `:global` attribute type natively includes:
- All standard HTML5 global attributes (such as `id`, `class`, `style`, etc.)
- All prefixed attributes starting with `phx-`, `aria-`, and `data-`

Explicitly listing these in the `:include` list is redundant. This change cleans up the attributes definitions to only specify custom, non-global attributes (such as `form`, `disabled`, and `multiple`) where required by the underlying HTML tags.

References:
- https://phoenix-live-view.hexdocs.pm/1.2.7/Phoenix.Component.html#module-global-attributes
- https://github.com/phoenixframework/phoenix_live_view/blob/v1.2.7/lib/phoenix_component/declarative.ex#L16-L120

### DIFF
--- a/lib/hexpm_web/components/buttons.ex
+++ b/lib/hexpm_web/components/buttons.ex
@@ -23,14 +23,9 @@ defmodule HexpmWeb.Components.Buttons do
   attr :full_width, :boolean, default: false
   attr :loading, :boolean, default: false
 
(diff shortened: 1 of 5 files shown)


ericmj:
💜

PR #1763, opened 2026-07-23
Remove obscured live package filtering on mobile

### LINKED ISSUE #1699: Proposal: Remove Obscured Live Package Filtering
Currently, on the `/packages` page on mobile devices, typing in the search input triggers a live query patch that refilters the package list on the page in real-time.

Image

However:
1. On physical mobile devices, the virtual keyboard fully covers/obscures the main content area of the page. The user cannot see the results updating, making the real-time feedback useless.
2. A new database query and live page update are triggered for every debounced keystroke, wasting resources.
3. On any other page (like `/` and `/packages/phoenix`), the search bar doesn't have live filtering, requiring the user to explicitly submit the form.

## Proposed Change

Remove real-time live-filtering from the package index page. Mobile and desktop search would behave the same, relying on explicit form submission. Desktop, however, has a live suggestion feature which would remain.

This change allows us to safely delete some code:

### 1. Mobile Search Inputs

In `lib/hexpm_web/components/navbar.ex`, we can remove the `phx-change` and `phx-debounce` attributes from the mobile search inputs:

* Mobile search bar: https://github.com/hexpm/hexpm/blob/1678160a5b8b61ae6f126cf9f6930e5fcd2318fb/lib/hexpm_web/components/navbar.ex#L240-L241
* Mobile menu search: https://github.com/hexpm/hexpm/blob/1678160a5b8b61ae6f126cf9f6930e5fcd2318fb/lib/hexpm_web/components/navbar.ex#L285-L286

### 2. Search Change Event Handler

In `lib/hexpm_web/live/package_live/index.ex`, the `search_change` event handler becomes dead code and can be removed completely: https://github.com/hexpm/hexpm/blob/1678160a5b8b61ae6f126cf9f6930e5fcd2318fb/lib/hexpm_web/live/package_live/index.ex#L366-L374

By removing this, we wouldn't need to handle events that don't change the assigns as in https://github.com/hexpm/hexpm/pull/1697/changes#diff-7bced4d349922f1979563c5b226cf6cd2059746b50385cff7ba0a9b5ff945fee.

rhcarvalho:
On mobile devices, typing in the search input triggered real-time package filtering on every debounced keystroke. However, physical mobile keyboards obscure the main content area, rendering the real-time feedback useless while sending unnecessary queries.

Mobile search now relies exclusively on explicit form submission.

## Summary

Closes #1699.

## Testing

Manually verified that search still works after explicit form submission.

### DIFF
--- a/lib/hexpm_web/components/navbar.ex
+++ b/lib/hexpm_web/components/navbar.ex
@@ -227,7 +227,6 @@ defmodule HexpmWeb.Components.Navbar do
           role="search"
           action={~p"/packages"}
           class="flex-1"
-          phx-change={if @live_search, do: "search_change"}
           phx-submit={@live_search && "search_submit"}
           phx-hook={@live_search && "FormSync"}
           data-sync-to={@live_search && "mobile-menu-search-form"}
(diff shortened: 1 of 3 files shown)


ericmj:
Thank you! 💜

PR #1766, opened 2026-07-24
Use search input type on mobile search inputs

rhcarvalho:
## Summary

Change the input type from `text` to `search` on the mobile search fields.

This prompts mobile browsers to natively render a clear ("X") button inside the search field when text is present, allowing mobile users to quickly clear their query and start over.

This also matches the desktop search form and is the semantic form field type in HTML5.

## Testing

Check on a mobile device/simulator that querying a new package is easier by clicking the standard "x" button to clear the current query before typing the new one, instead of having to deal with text selection or soft keyboard backspace.

### Before

hex.pm before

### After

hex.pm after

### DIFF
--- a/lib/hexpm_web/components/navbar.ex
+++ b/lib/hexpm_web/components/navbar.ex
@@ -238,7 +238,7 @@ defmodule HexpmWeb.Components.Navbar do
             <input
               id="mobile-search-input"
               name="search"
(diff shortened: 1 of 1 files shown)


ericmj:
Thanks for all your testing and improvements of the mobile experience!

PR #1785, opened 2026-07-31
feat: trusted publisher

yordis:


yordis:
🤦🏻 OMG I meant to create a PR against my own repo to code review myself and somehow I ended up messing it up;

I am still in active development and I am holding the PR Description after I finish 😭 Sorry....

---

By now, do you prefer I close it or just continue the work and leave it here for visibility.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
+# Trusted Publishers: npm vs PyPI (and what it means for Hex)
```
maennchen, line 1: This document should be re-homed to https://github.com/hexpm/specifications before merge.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
...
+
+Both require the same core identity and treat environment as optional. npm
+constrains the workflow to a filename; PyPI does the same.
+
+| Field | npm | PyPI |
+| --- | --- | --- |
+| Owner / org | Required | Required |
+| Repository | Required | Required |
+| Workflow | Required, **filename only** (e.g. `publish.yml`), must be in `.github/workflows/` | Required, filename (e.g. `release.yml`) |
+| Environment | Optional | Optional, **strongly recommended** (enables manual-approval gate) |
```
maennchen, line 102: Since we're starting this now, I would require this. For example `azure/login` also does this.
Requires a bit more safety from the start.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
...
+| Allowed actions | Required: `npm publish` and/or `npm stage publish` | n/a |
```
maennchen, line 103: I think it would be beneficial to allow reading as well. For example a GitHub Action could authenticate a `mix deps.get` from provate org packages instead of having a long-lived token stored anywhere.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
...
+
+### 2.3 Package <-> publisher cardinality
+
+| | npm | PyPI |
+| --- | --- | --- |
+| Model | **One** trusted publisher per package at a time (editable) | **Many-to-many** |
+| Rationale | Simplicity | One publisher -> many projects (monorepos, tandem releases); one project -> many publishers (per-OS/arch binary wheels) |
```
maennchen, line 110: One publisher -> many projects should be supported for monorepos. For example RabbitMQ publishes multiple packages from one big repo. Not sure if we need the inverse.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
...
+| | npm | PyPI |
+| --- | --- | --- |
+| Pending publishers | Not documented (package must already exist) | Yes - configure a "pending" publisher under your account; first successful publish creates the project. Name is not reserved until first publish; if someone else registers it first, the pending publisher is invalidated |
+
+### 2.6 Provenance / attestations
+
+| | npm | PyPI |
+| --- | --- | --- |
+| Automatic provenance | Yes, by default when publishing via OIDC from GitHub/GitLab, for **public** repo + **public** package (not CircleCI, not private repos) | Digital attestations (PEP 740) with Sigstore |
+| Opt out | `NPM_CONFIG_PROVENANCE=false`, `.npmrc`, or `publishConfig.provenance` | n/a |
```
maennchen, line 150: This is a bigger topic and should be excluded from the current scope. I however like the idea that we store the claims of the OIDC token and attach that to a publish. This way we can show in the UI that this was published by trusted publishing and what workflow it was.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,204 @@
...
+
+### 2.7 Hardening around tokens
+
+| | npm | PyPI |
+| --- | --- | --- |
+| Disallow long-lived tokens | Package setting: "Require 2FA and disallow tokens" (does not affect OIDC) | Recommends removing manual tokens after switching |
+| Staged publishing | Yes - `npm stage publish` can require a maintainer to approve with 2FA before the release goes public | No direct equivalent |
```
maennchen, line 157: That would be super nice and we should think about supporting it long term. We could also use that in normal `mix hex.publish`. That way we could prompt for 2FA in the browser instead of entering an OTP token in the CLI. That would also allow WebAuthn based MFA.

lib/hexpm/trusted_publishers/provider/github.ex
```diff
@@ -0,0 +1,169 @@
...
+    end
+  end
+
+  defp fetch_repository_id(repository, owner) do
+    full_name =
+      if String.contains?(repository, "/") do
+        repository
+      else
+        "#{owner}/#{repository}"
+      end
```
maennchen, line 107: Why would this sometimes contain / not contain the owner?
Could the `owner_part` be out of sync with `owner`?

lib/hexpm/trusted_publishers/provider/github.ex
```diff
@@ -0,0 +1,169 @@
...
+
+    case String.split(full_name, "/") do
+      [owner_part, name] ->
+        if valid_repo_name?(name) do
+          path = "/repos/#{URI.encode(owner_part)}/#{URI.encode(name)}"
+
+          case github_get(path) do
```
maennchen, line 114: For private packages we probably need to provide some auth header to not get a 404.

lib/hexpm/trusted_publishers/provider/github.ex
```diff
@@ -0,0 +1,169 @@
...
+            {:ok, %{"id" => id}} when is_integer(id) -> {:ok, id}
+            {:ok, _} -> {:error, :repository_not_found}
+            {:error, :not_found} -> {:error, :repository_not_found}
+            {:error, reason} -> {:error, reason}
```
maennchen, line 118: Why do we not extract the owner id and repository id at the same time? It's present on the repo response.

lib/hexpm/trusted_publishers/oidc.ex
```diff
@@ -0,0 +1,271 @@
...
+  """
+  def verify(token, issuer) when is_binary(token) and is_binary(issuer) do
+    with {:ok, header} <- peek_header(token),
+         :ok <- validate_alg(header),
+         {:ok, jwks} <- get_jwks(issuer),
+         {:ok, claims} <- verify_signature(token, jwks, issuer),
+         :ok <- validate_claims(claims, issuer) do
+      {:ok, claims}
+    end
+  end
```
maennchen, line 42: Can we run this against `Oidcc.Token.validate_jwt/3` isntead of directly Joken?

lib/hexpm/trusted_publishers/provider.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Hexpm.TrustedPublishers.Provider do
```
maennchen, line 1: This interface is not very generic. We should either implement against GitHub directly in the first stage and not try to make it generic or find a good abstraction.

lib/hexpm/trusted_publishers/trusted_publisher.ex
```diff
@@ -0,0 +1,126 @@
...
+  defp normalize_optional(value), do: value |> String.trim() |> String.downcase()
+
+  defp normalize_workflow(nil), do: nil
+
+  defp normalize_workflow(workflow) do
+    workflow
+    |> String.trim()
+    |> Path.basename()
+    |> String.downcase()
+  end
```
maennchen, line 119: That sounds a bit dangerous. For example, would git allow having two workflows with the same name and just different casing?

Auth attributes should be exact and allow no wiggle room.

docs/trusted_publishers_comparison.md
yordis: Both directions are supported: publisher configs are per-package rows with uniqueness scoped to the package, so the same repository + workflow can back any number of packages (the RabbitMQ case), and a package can hold several publishers.

Covered by tests at both the context and HTTP publish level, and the docs now note that a monorepo workflow mints once per package with a fresh OIDC token each time, since a jti can only be minted once.

lib/hexpm/trusted_publishers/oidc.ex
yordis: Done. `Oidcc.Token.validate_jwt/3` now handles signature, `iss`, `aud`, `exp`, and `nbf`. What stayed local is only what it does not cover: rejecting `none` and HMAC before any key work, an upper bound on `iat`, and requiring `jti` for replay protection. Joken is down to unverified peeking.

Two things need your call:

- GitHub's discovery document has no `authorization_endpoint`, so `Oidcc.ProviderConfiguration.decode_configuration/1` cannot decode it. The configuration is built from the issuer and allowed algorithms directly, same shape as `harden_configuration/1` in the SSO oidcc module.
- `oidcc` reads clock skew from the global `config :oidcc, max_clock_skew`, which is unset here, so `exp` and `nbf` are now checked with zero skew. Setting it would also loosen SSO id_token validation, so I kept the 60s tolerance for `iat` only. Do you want skew on `exp` and `nbf` too?

lib/hexpm/trusted_publishers/provider.ex
yordis: You are right, and part of it was dead weight. `required_claims/0` and `supported_claims/0` had no callers at all, so they are gone along with `known_providers/0`. Provider and issuer lookup now derive from one list of implementations instead of two hardcoded maps.

What is left is genuinely GitHub-shaped: the schema columns (`repository_owner`, `repository`, `workflow`, `environment`) and the `immutable_ids` keys. Making those generic means a per-provider config payload plus a migration.

Should stage one stay GitHub-only in the schema and grow the abstraction when a second provider lands, or does the config payload go in now while the table is still empty?

lib/hexpm/trusted_publishers/trusted_publisher.ex
yordis: Fixed. Workflow and environment keep their casing through the changeset and are compared exactly, so `Release.yml` no longer satisfies a publisher configured for `release.yml`. Workflow matching also requires a value on both sides now, so two `nil`s cannot match each other.

Owner and repository are a deliberate exception: downcased and compared case-insensitively, because the GitHub namespace is itself case-insensitive, and pinned by their immutable GitHub ids on top of that.

Worth flagging that this makes us stricter than PyPI on `environment`, whose `_check_environment` compares `ground_truth.lower() == signed_claim.lower()`. I went exact on the grounds that this is auth configuration, so the looser side should be the one that has to be argued for. The cost is that a publisher configured `Production` against an environment named `production` is accepted at create and then never matches, surfacing only at publish time. I have not verified whether GitHub even permits two environments in one repository differing only by case; if it does not, case-insensitive would lose nothing. One `String.downcase/1` on both sides either way.

Is the owner/repository split what you want, or would you rather those be exact as well?

lib/hexpm/trusted_publishers/provider/github.ex
yordis: Done. `resolve_immutable_ids/1` makes a single `GET /repos/:owner/:name` call and reads `id` and `owner.id` from that one response.

That collapsed the error surface too. A 404 is `:repository_not_found` and surfaces as a 422, because with one call it cannot tell "no such owner" from "no such repository" from "private repository we cannot see". A 200 missing `id` or `owner.id` is GitHub violating its own contract rather than a missing owner, so it is `:invalid_github_response` and surfaces as a 503.

Unrelated fix while I was in there: the old catch-all turned every non-404 failure (rate limit, SSO-blocked org, GitHub 5xx, socket timeout) into a 422 on `repository_owner`, which tells the client its input is wrong, and clients do not retry a 422. Those are 503 with a retryable message now.

lib/hexpm/trusted_publishers/provider/github.ex
yordis: The real problem was upstream of this line: `repository` could be stored either bare (`widget`) or owner-qualified (`acme/widget`), so the provider had to cope with both and the two representations could drift.

The changeset now canonicalizes `repository` to `owner/name` and rejects anything whose owner segment does not match `repository_owner`, so there is one representation in the database. The owner-prefixing moved out of the context into the changeset, and `owner_part` is gone.

lib/hexpm/trusted_publishers/provider/github.ex
yordis: Private repositories work now, though not through the auth header, because that header cannot solve it. `github_token/0` falls back to `HEXPM_DOCS_GITHUB_TOKEN`, which is hex.pm's own credential and has no access to somebody else's private repository, and the user-facing GitHub OAuth scope is `user:email`, so there is no per-user token to borrow either. Reading arbitrary private repositories would mean a GitHub App that users install on their org.

For the owner id it turned out to be unnecessary: `GET /users/:owner` is public even when the owner's repositories are not, so resolution falls back to it when `/repos/:owner/:name` answers 404.

The repository id is different, since we cannot read it for a repository we cannot see. The client supplies `repository_id` (`gh api repos/OWNER/NAME --jq .id`), and create fails with a 422 on that field when we cannot resolve it and it was not sent. Both ids are always pinned, `repository_id` is `null: false`, and matching fails closed when either side is missing.

I considered dropping `repository_id` to match PyPI, which stores `repository_owner_id` and has no `repository_id` column, and decided against it. Owner-id pinning does not cover the org-internal case: if a trusted repository is deleted, any member with plain repository-creation rights (much more widely granted than admin or publish) can recreate the name, add the configured workflow, and mint. A recreated repository is a different repository, so there is deliberately no trust-on-first-use window where an unpinned publisher adopts whatever id shows up first.

One thing I noticed reading PyPI's implementation, outside this thread but worth your call: their `__required_verifiable_claims__` refuses tokens where `event_name` is `pull_request_target`, and we do not look at `event_name` at all. It takes the package's own workflow to be configured that way, so it is defense in depth rather than a live hole, but it is one clause. Want it in this PR or a follow-up?

docs/trusted_publishers_comparison.md
yordis: Answered in https://github.com/hexpm/hexpm/pull/1785#discussion_r3717208926: both directions work. A package can have several publishers, and the same repository and workflow can back several packages as separate rows, which is the monorepo case. Both are covered by tests, including through the HTTP flow.

The one constraint worth calling out, and it is in the docs: when a single workflow run publishes several packages, it has to mint once per package and request a fresh OIDC token for each mint, because a token's `jti` is single use.

Anything else you want spelled out here, or can this one be resolved?

docs/trusted_publishers_comparison.md
yordis: Implemented in 383166f0 and 1a36d316. An allowlisted snapshot of the verified claims (repository, workflow ref, commit sha, run id, actor, event name, and similar) is attached to the minted token and carried onto the release published with that token. The release API exposes it as `oidc_claims`, so a consumer can see that a release came from trusted publishing and which workflow produced it.

Full provenance and attestations are still out of scope for this PR and listed under Limitations.

docs/trusted_publishers_comparison.md
yordis: I would rather keep it optional, because neither reference registry requires it and our storage model already matches PyPI's.

PyPI keeps `environment` in `__optional_verifiable_claims__`, not `__required_verifiable_claims__`, with the column `nullable=False` and `""` meaning "no constraint", and `_check_environment` short-circuits to true on the empty string. That is the same shape as our `null: false, default: ""` plus the empty-string clause in `environment_matches?/2`. npm lists it as "Environment name (optional)". PyPI's own docs call it "optional but strongly recommended".

The reason I do not think requiring it buys anything: an environment is only valuable because it is where GitHub lets you attach deployment protection rules (required reviewers, wait timers, branch restrictions). Forcing someone to type a name for an environment with no protection rules configured gives them the ceremony without the gate. So I have changed the docs from "Optional" to "optional but strongly recommended" with that reason spelled out, rather than changing the schema.

Happy to flip it to required if you disagree, and you are right that now is the cheap moment since no rows exist.

docs/trusted_publishers_comparison.md
```diff
@@ -0,0 +1,213 @@
...
+## 4. Where they differ (Hex has to choose)
+
+| Decision | npm | PyPI | Hex plan |
+| --- | --- | --- | --- |
+| Cardinality | One per package | Many-to-many | `package has_many :trusted_publishers` (covers monorepo + per-platform without a join table) |
+| Immutable ID | Exact string match only | Resolve + store `repository_owner_id` | Follow PyPI: resolve at config time, check at mint time |
+| Config-time verification | None | Yes (implied by ID resolution) | Follow PyPI: resolve immutable IDs when the publisher is created |
+| Pending publishers | No | Yes | Defer to a later phase |
+| Provenance | Auto | Attestations | Defer (design schema to allow it later) |
+| Mint endpoint | OIDC token exchange endpoint | `/_/oidc/mint-token` | Dedicated `POST /api/oidc/mint-token` |
```
ericmj: Should we use `POST /api/oauth/token` with a new `urn:ietf:params:oauth:grant-type:jwt-bearer` grant instead?

lib/hexpm/accounts/audit_log.ex
```diff
@@ -19,7 +19,7 @@ defmodule Hexpm.Accounts.AuditLog do
   end
 
   def build(%{user: nil} = audit_data, action, params)
-      when action in ~w(password.reset.init password.reset.finish user_provider.create user.delete) do
+      when action in ~w(password.reset.init password.reset.finish user_provider.create user.delete trusted_publisher.mint) do
```
ericmj: This can't be an audit log entry if we extend this to fetching packages. It would create an entry on every CI run.

lib/hexpm_web/templates/docs/trusted_publishers.html.md
```diff
@@ -0,0 +1,236 @@
...
+          BODY=$(jq -n \
+            --arg token "$OIDC_TOKEN" \
+            --arg package "PACKAGE" \
+            '{token: $token, package: $package}')
+
+          MINT=$(curl -fsS -X POST "$HEX_API_URL/oidc/mint-token" \
+            -H "content-type: application/json" \
+            -d "$BODY")
+
+          export HEX_API_KEY=$(printf '%s' "$MINT" | jq -r .token)
```
ericmj: This could be part of `erlef/setup-beam` or a new action and/or a mix task?

lib/hexpm_web/templates/docs/trusted_publishers.html.md
```diff
@@ -0,0 +1,236 @@
...
+* Configuring or removing a publisher requires full package ownership and `api:write`. Two-factor authentication is not required for CI mint/publish because the trusted-publisher grant has no interactive user.
+* Hex records an allowlisted snapshot of the verified OIDC claims (repository, workflow ref, commit SHA, run id, and similar) on the minted token and attaches it to the release published with that token. The release API exposes this as `oidc_claims`, so consumers can see that a release was published via trusted publishing and which workflow produced it.
+
+### Limitations
+
+* GitHub Actions only in this release. GitLab, CircleCI, and custom issuers are not supported yet.
+* No dashboard UI; use the management API.
+* No Mix / Rebar3 built-in trusted-publisher commands yet. Clients should request the CI OIDC token with audience from `/api/oidc/audience`, call `/api/oidc/mint-token`, and publish with the returned bearer token.
+* Cannot create a package or land the first release from CI. Publish once manually, then attach a trusted publisher.
+* Provenance / attestations, pending publishers, and a package setting to disallow long-lived tokens are deferred.
```
ericmj: This doesn't really fit in user facing docs. It should move to PR description or comment.

lib/hexpm_web/controllers/api/trusted_publisher_controller.ex
```diff
@@ -0,0 +1,160 @@
+defmodule HexpmWeb.API.TrustedPublisherController do
```
ericmj: I don't think we want to expose this through the API. We need dashboard UI for this instead.

lib/hexpm_web/controllers/api/oidc_controller.ex
```diff
@@ -0,0 +1,125 @@
...
+    else
+      conn
+      |> put_status(404)
+      |> render(:error, error_type: :not_found, description: "Not found")
+      |> halt()
+    end
+  end
+
+  defp mint_rate_limit(conn, _opts) do
+    case Attack.trusted_publisher_mint_ip_throttle(conn.remote_ip) do
```
ericmj: Not sure about this, maybe we should only rate limit failed minting attempts. This would likely run into rate limits when running from a shared pool of IPs from CI workers.

In fact, these routes should probably be exempt from the normal API rate limiting.

priv/repo/migrations/20260730213924_seed_trusted_publisher_oauth_client.exs
```diff
@@ -0,0 +1,28 @@
+defmodule Hexpm.Repo.Migrations.SeedTrustedPublisherOauthClient do
+  use Ecto.Migration
+
+  @client_id "a1111111-1111-4111-8111-111111111111"
```
ericmj: What is this for?

priv/repo/seeds.exs
```diff
@@ -33,6 +33,19 @@ Hexpm.Repo.transaction(fn ->
     ]
   )
 
+  trusted_publisher_client_id = "a1111111-1111-4111-8111-111111111111"
+
+  unless Hexpm.Repo.get(Hexpm.OAuth.Client, trusted_publisher_client_id) do
```
ericmj: Why is this conditional?

priv/repo/migrations/20260804090001_add_trusted_publisher_to_releases.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Hexpm.Repo.Migrations.AddTrustedPublisherToReleases do
```
ericmj: Please combine all migrations into a single one.

lib/hexpm_web/auth_helpers.ex
```diff
@@ -54,6 +61,34 @@ defmodule HexpmWeb.AuthHelpers do
...
+
+    cond do
+      not Keyword.get(opts, :allow_trusted_publisher, false) ->
+        error(conn, {:error, :auth})
+
+      not verify_permissions?(conn, auth_credential, domains) ->
+        error(conn, {:error, :domain})
+
+      true ->
+        case package_owner(conn, trusted_publisher, opts) do
```
ericmj: This crashes for packages in an organization repository. `package_owner/3` goes through `package_sso_enforced`, and `Hexpm.Accounts.SSO.Enforcement.check/4` has no clause for a `%TrustedPublisher{}` principal, so `POST /api/repos/:org/publish` with one of these tokens raises `FunctionClauseError`. Docs uploads take the same path, and none of the tests publish to an organization repository.

lib/hexpm/trusted_publishers/provider/github.ex
```diff
@@ -0,0 +1,214 @@
...
+  defp workflow_matches?(%{workflow: expected}, actual)
+       when is_binary(expected) and is_binary(actual),
+       do: expected == actual
+
+  defp workflow_matches?(_publisher, _actual), do: false
+
+  defp environment_matches?(%{environment: env}, _token_environment) when env in [nil, ""],
+    do: true
+
+  defp environment_matches?(%{environment: expected}, actual), do: expected == actual
```
ericmj: GitHub environment names aren't case sensitive (https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/manage-environments), so a publisher configured with `Production` doesn't match an environment GitHub has as `production`, even though GitHub treats them as the same environment. This should compare them case-insensitively, like PyPI's `_check_environment` does (https://github.com/pypi/warehouse/blob/main/warehouse/oidc/models/github.py), and the docs that say it's case-sensitive need the same change.

lib/hexpm/trusted_publishers/provider/github.ex
```diff
@@ -0,0 +1,214 @@
...
+      fetch_repository_ids(owner, name)
+    end
+  end
+
+  # The GitHub namespace is case-insensitive, so owner and repository are
+  # compared case-insensitively and pinned by their immutable ids. Workflow and
+  # environment are matched exactly: git paths are case-sensitive, so
+  # `Release.yml` must not satisfy a publisher configured for `release.yml`.
+  @impl true
+  def match?(%TrustedPublisher{} = publisher, claims) when is_map(claims) do
```
ericmj: This should reject tokens where `event_name` is `pull_request_target`, which PyPI does (https://github.com/pypi/warehouse/blob/main/warehouse/oidc/models/github.py). That event runs with the base repository's permissions, so a workflow that checks out the pull request's head lets code from a fork request an OIDC token.

lib/hexpm/trusted_publishers/oidc.ex
```diff
@@ -0,0 +1,290 @@
...
+    else
+      nil -> {:error, :jwks_uri_missing}
+      {:error, reason} -> {:error, reason}
+      _ -> {:error, :discovery_failed}
+    end
+  end
+
+  defp put_jwks_cache(issuer, jwks, expires_at) do
+    case :persistent_term.get({__MODULE__, :jwks, issuer}, :miss) do
+      {:ok, ^jwks, _expires_at} ->
```
ericmj: When the refetched JWKS is the same as the cached one, this keeps the old `expires_at`, so the entry stays expired and every exchange after the first hour fetches discovery and the JWKS from GitHub again. GitHub sends `max-age=3600` on the JWKS. Line 102 already writes to `:persistent_term` on every fetch, so this can always store the new expiry.

priv/repo/migrations/20260730213923_add_trusted_publisher_to_oauth_tokens.exs
```diff
@@ -0,0 +1,45 @@
...
+    create_if_not_exists unique_index(
+                           :oauth_tokens,
+                           [:grant_reference, :client_id],
+                           where: "grant_type = 'trusted_publisher' AND grant_reference IS NOT NULL",
+                           name: :oauth_tokens_trusted_publisher_grant_reference_client_id_index
+                         )
+
+    drop_if_exists constraint(:oauth_tokens, :user_or_organization_required)
+
+    create constraint(:oauth_tokens, :user_or_organization_or_trusted_publisher_required,
```
ericmj: Adding a validated check constraint takes an ACCESS EXCLUSIVE lock on `oauth_tokens` and scans every row (https://www.postgresql.org/docs/current/sql-altertable.html), which blocks every OAuth token lookup until it finishes. `validate: false` followed by a separate `VALIDATE CONSTRAINT` avoids that. The unique index above should be built with `concurrently: true` in a migration with `@disable_ddl_transaction` and `@disable_migration_lock`, like `20260417120000_optimize_audit_logs_indexes.exs`.

priv/repo/migrations/20260730213922_create_trusted_publishers.exs
```diff
@@ -0,0 +1,28 @@
...
+      add :repository_id, :string, null: false
+      add :repository, :string, null: false
+      add :workflow, :string, null: false
+      add :environment, :string, null: false, default: ""
+
+      timestamps()
+    end
+
+    create index(:trusted_publishers, [:package_id])
+    create index(:trusted_publishers, [:provider, :repository, :workflow])
```
ericmj: Both of these can go. The unique index below leads with `package_id`, which covers the lookups by `package_id` and by `package_id` and `provider`, and nothing queries by `[:provider, :repository, :workflow]`.

lib/hexpm/trusted_publishers.ex
```diff
@@ -0,0 +1,292 @@
...
+  end
+
+  def get(id) do
+    case parse_id(id) do
+      {:ok, id} -> Repo.get(TrustedPublisher, id)
+      :error -> nil
+    end
+  end
+
+  def create(%Package{} = package, params, audit: audit_data) do
```
ericmj: Adding or removing a trusted publisher should email every package owner, like adding an owner does with `Emails.owner_added`. A publisher lets anyone who can run that workflow publish the package.

PR #1786, opened 2026-07-31
Group Dependabot Updates

maennchen:
## Summary

Reduce Weekly Noise by Dependabot Updates

Downside: If there's a single dependency that does not work, it will fail the whole group. => Manual Action needed.

## Testing

We'll see it next week.


ericmj:
Thank you! 💜

PR #1806, opened 2026-08-05
Scan published packages for leaked credentials

ericmj:
Packages get published with credentials in them: a `.env` that slipped past `:files`, a `config/prod.secret.exs`, an AWS key in a test fixture, a Phoenix `secret_key_base`. Once a release is on repo.hex.pm it's mirrored and cached, so the only remedy is rotating the credential, and today nobody tells the owner.

This adds the notification half of what GitHub does: scan what was published, email the owners so they can rotate, take no other action. The publish isn't blocked, nothing is revoked, findings aren't listed publicly.

The scan runs as its own Oban job at priority 5, enqueued from the same SQS ObjectCreated event that drives preview processing. It calls the same `download_and_unpack!/3` the preview path uses, but into its own temp directory, so it pays for a second download and unpack. That buys independent retries and lets it yield to preview and diff work rather than sharing their attempts. Rules are the gitleaks ruleset (222 rules, MIT), vendored under `data/secret_scan/` and parsed at compile time. `mix hexpm.secret_scan.refresh` applies one rewrite on the way in: gitleaks allowlists example keys by spelling them out, so the last byte of each becomes a character class. Same regex, but our copy holds no literal credential for GitHub's own secret scanning to alert on. A keyword prefilter runs before any regex, which is the difference between 2 s and 146 ms per 256 KB window. Files are scanned in a streaming window (256 KB with a 64 KB carry) so peak memory stays bounded regardless of file size.

Only a fingerprint (HMAC-SHA256 with the hexpm secret) and a masked preview are stored, never the raw value. Recipients follow ownership: package owners, org admins when the package is private or owned by an organization, and the publisher. In production the scanner records and emails only on a curated allowlist of high-confidence rules (checksummed or hard-prefixed provider keys, plus verified Hex keys); the full ruleset runs only in the offline corpus tool.

Publishers can exclude paths with `package: [secret_scan: [ignore: ["test/fixtures/**"]]]` in `mix.exs`. That relies on the companion hex change to carry the key into the tarball, and the field is specified in the specifications repo.

Companion PRs: hexpm/hex#1221, hexpm/specifications#77


lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,448 @@
...
+      file_paths
+      |> Task.async_stream(&scan_file(dir, &1, opts),
+        max_concurrency: System.schedulers_online(),
+        timeout: file_timeout(),
+        on_timeout: :kill_task
+      )
+      # Lazily: `Enum.zip/2` would drain the whole stream before the deadline
+      # below ever got to look at the clock, which is the difference between a
+      # minute of CPU and however long the release feels like taking.
+      |> Stream.zip(file_paths)
```
maennchen: Since this is `ordered: true`, if one file does not scan, we will never emit any later files.

We could instead return `{path, result}` from `scan_file` and provide `zip_input_on_exit: true, ordered: false`

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+        ordered: false,
+        zip_input_on_exit: true
+      )
+      |> Enum.reduce_while(empty, fn result, state ->
+        if System.monotonic_time(:millisecond) > deadline do
+          {:halt, %{state | halted: true}}
+        else
+          {:cont, accumulate(result, state)}
+        end
+      end)
```
maennchen: Is it problematic that this can go over the release timeout?

If a file that times out is consumed just before the deadline, it could essentially take `release timeout + file timeout` time.

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+  def scan_file(dir, path, opts \\ []) do
+    scan(path, fn -> dir |> Path.join(path) |> stream_windows() end, opts)
+  rescue
+    # A task raising here would leave the stream as an exit, which the caller's
+    # rescue cannot catch, so one unreadable file would kill the whole release's
+    # scan and take the preview job with it.
+    exception ->
+      Logger.warning("Secret scan could not read #{path}: #{Exception.message(exception)}")
+      []
+  end
```
maennchen, line 240: This currently assumes that the path is safe. And since we have `hex_tarball` doing checks, that's probably right. But I would make sure the path is really safe before we read it.

```suggestion
  def scan_file(dir, path, opts \\ []) do
    with {:ok, path} <- Path.safe_relative(path, dir),
         true <- File.regular?(Path.join(dir, path)) do
      scan_file(dir, path, opts)
    else
      :error ->
        Logger.warning("Secret scan could not read #{path}: not a safe relative path")
        []
      false ->
        Logger.warning("Secret scan could not read #{path}: not a regular file")
        []
    end
  rescue
    # A task raising here would leave the stream as an exit, which the caller's
    # rescue cannot catch, so one unreadable file would kill the whole release's
    # scan and take the preview job with it.
    exception ->
      Logger.warning("Secret scan could not read #{path}: #{Exception.message(exception)}")
      []
  end
```

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+  # `**` spans directory separators, `*` stays within a segment, `?` is one
+  # non-separator character. A malformed glob is dropped rather than crashing
+  # the scan.
+  defp compile_globs(globs), do: Enum.flat_map(globs, &compile_glob/1)
+
+  defp compile_glob(glob) when is_binary(glob) do
+    pattern =
+      glob
+      |> String.split(~r/(\*\*|\*|\?)/, include_captures: true, trim: true)
+      |> Enum.map_join(fn
```
maennchen: This would not have the glob `src/**/foo.ex` match `src/foo.ex`.

```suggestion
      |> String.split(~r{(/\*\*/|\*\*|\*|\?)}, include_captures: true, trim: true)
      |> Enum.map_join(fn
        "/**/" -> "(?:/.*/|/)"
```

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+  # table cannot test a candidate against a finding or brute-force a low-entropy
+  # one.
+  defp fingerprint(value) do
+    :crypto.mac(:hmac, :sha256, Application.fetch_env!(:hexpm, :secret), value)
+  end
+
+  # gitleaks honours an inline `gitleaks:allow` on the line, which is how a
+  # project marks a deliberate fixture. A package that vendors an annotated
+  # test file should not earn its owner a security email.
+  defp annotated_allowed?(line), do: String.contains?(line, @inline_allow)
```
maennchen, line 502: I'm kind of wondering if we should publish the scanner itself as a package as well so that a project can check against leaks locally as well. Especially for adding ignores, possibly running the scanner in a CI etc.

lib/hexpm/secret_scan/scanner.ex
maennchen: I embarked on a little side quest to make this simpler in the future: https://github.com/elixir-lang/elixir/pull/15720

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+    # unsalted sha256 of the value, so on a ten character secret the old fixed
+    # four-and-four left two unknown characters and the hash to check them
+    # against, which is not redaction.
+    keep = if size < 12, do: 0, else: min(4, div(size, 6))
+
+    if keep == 0 do
+      String.duplicate("*", min(size, @preview_stars))
+    else
+      binary_part(secret, 0, keep) <>
+        String.duplicate("*", @preview_stars) <> binary_part(secret, size - keep, keep)
```
maennchen: Since we run the scanners on binary files, they could theoretically include non printable characters. That will trigger `UnicodeConversionError` in email rendering. So we should make sure that our stored values are printable.

```suggestion
      start = secret |> binary_part(0, keep) |> String.replace_invalid()
      middle = String.duplicate("*", @preview_stars)
      end = secret |> binary_part(size - keep, keep) |> String.replace_invalid()
      start <> middle <> end
```

lib/hexpm/secret_scan/finding.ex
```diff
@@ -0,0 +1,62 @@
...
+  def location(%__MODULE__{} = finding), do: location(Map.from_struct(finding))
+
+  def location(%{file_path: path, line: line}) when is_integer(line) and line > 0 do
+    "#{display_path(path)}:#{line}"
+  end
+
+  def location(%{file_path: path, byte_offset: offset}), do: "#{display_path(path)}@#{offset}"
+
+  defp display_path(path) do
+    path
```
maennchen, line 58: Could theoretically also be non UTF-8

```suggestion
    path
    |> String.replace_invalid()
```

priv/repo/migrations/20260802170000_create_secret_scans.exs
```diff
@@ -0,0 +1,43 @@
...
+      add :file_path, :text, null: false
+      add :line, :integer, null: false
+      add :byte_offset, :integer, null: false
+      add :fingerprint, :binary, null: false
+      add :preview, :text, null: false
+
+      timestamps(type: :utc_datetime_usec, updated_at: false)
+    end
+
+    create unique_index(:secret_findings, [:release_id, :fingerprint])
```
maennchen: If you have the same secret in multiple files, would you get a conflict here?

lib/hexpm/secret_scan/secret_scan.ex
```diff
@@ -0,0 +1,360 @@
...
+          tarball_checksum: tarball_checksum,
+          inserted_at: now
+        })
+      end)
+
+    {_count, inserted} =
+      repo.insert_all(Finding, entries,
+        on_conflict: :nothing,
+        conflict_target: [:release_id, :fingerprint],
+        returning: true
```
maennchen, line 201: If we rescan on a new tarball, the findings would keep their old `tarball_checksum`.

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+          if MapSet.member?(seen, key) do
+            {found, seen}
+          else
+            finding = finding(rule, match, content, path, offset, newlines_before)
+            {found ++ List.wrap(finding), MapSet.put(seen, key)}
+          end
+
+        # One over the cap, so a window producing exactly the cap stays
+        # distinguishable from one that overflowed it.
+        if length(found) > @max_findings do
```
maennchen, line 409: If I prepare a file with lots of fake tokens and very little entropy, I could grow seen a lot while staying under the max finding ceiling.

Should we also enforce a max seen or something like that?

lib/hexpm/secret_scan/worker.ex
```diff
@@ -0,0 +1,24 @@
...
+  reliability to the scanner's. At a lower priority than preview and diff, it
+  waits behind the user-facing work. It is idempotent, so a transient failure
+  retries rather than being dropped.
+  """
+
+  use Oban.Worker,
+    queue: :heavy,
+    priority: 5,
+    max_attempts: 5,
+    unique: [period: :infinity, states: :incomplete, fields: [:worker, :args]]
```
maennchen, line 17: Args receives a full S3 Event. For uniqueness, only the key is relevant.

```suggestion
    unique: [period: :infinity, states: :incomplete, fields: [:worker, :args], keys: ["key"]]
```

lib/hexpm/secret_scan/scanner.ex
```diff
@@ -0,0 +1,449 @@
...
+  Masks a secret for display, keeping enough of it to be recognisable.
+
+  The raw value is never stored or mailed — only this and a sha256 of it. The
+  mask is a fixed width rather than the secret's length, so a matched PEM block
+  doesn't become a three kilobyte row of asterisks.
+  """
+  def redact(secret) do
+    size = byte_size(secret)
+    # Never show more than a sixth from each end. The row also carries an
+    # unsalted sha256 of the value, so on a ten character secret the old fixed
```
maennchen:
```suggestion
    # Never show more than a sixth from each end. The row also carries a
    # sha256 HMAC of the value, so on a ten character secret the old fixed
```

lib/hexpm/secret_scan/scanner.ex
Ch4s3: I think this would be really nice.

priv/repo/migrations/20260802170000_create_secret_scans.exs
ericmj: Added multiple locations per credential (up to 10).

lib/hexpm/secret_scan/secret_scan.ex
ericmj: Retained rows now upsert tarball_checksum, rule, line and byte_offset, and freshness comes from the fingerprints on the release before the scan, so a rescan still doesn't re-mail

lib/hexpm/secret_scan/scanner.ex
ericmj: Fixed by pruning rather than a cap. seen only has to remember the overlap the next window rescans, so keys the windows have moved past get dropped. That bounds it by the matches in one window, the same order as the list Regex.scan/3 already builds for one rule, with nothing silently truncated.

lib/hexpm/secret_scan/scanner.ex
ericmj: Yes, and it's fixed in cded5c1. The stream is consumed in its own process now with the deadline as a receive timeout in the caller, so the budget is a ceiling and whatever is still running gets killed rather than left to burn CPU to completion. It also contains a crash: Task.async_stream propagates a task that raises or exits to whoever consumes the stream, so that used to take the release's scan and the job with it, and it's a monitor message now.

lib/hexpm/secret_scan/scanner.ex
ericmj: Done: https://hex.pm/packages/secret_scan

maennchen (approved):
Beautiful, ship it 😃

PR #1830, opened 2026-08-13
Pin Req to 0.6.3

ericmj:
Pin Req to 0.6.3 because Req 0.7 rewrites GET requests with an empty body to POST. ExAws signs these requests as GET, so S3 rejects them with `SignatureDoesNotMatch`.

Track the upstream incompatibility in https://github.com/ex-aws/ex_aws/issues/1246. The pin can be removed after a compatible ExAws release.

Verified with `mix format --check-formatted`, a local ExAws adapter regression check, and `mix test` (2,669 tests and 28 properties).


wojtekmach:
Oops, sorry about that. It's a req bug, I'll fix it upstream.

PR #1831, opened 2026-08-13
Update Lumis

leandrocp:
## Summary

Ref
https://github.com/leandrocp/lumis/blob/main/packages/elixir/lumis/CHANGELOG.md#070-2026-08-13

The main change is that languages (parsers) are now loaded on demand.

`Lumis.Languages.async_load/1` downloads a WASM file from a trusted source, compiles, and loads to warm-up languages, so I added the most popular ones but the rest are loaded lazily once they are requested and stay loaded on next requests. The main benefit is a way smaller package size.

Also revert changes on Dockerfile since it doesn't need to compile from source anymore.

## Testing

You can add code blocks to any project readme, for example:

lumis_0-7


ericmj:
- Can you look at the docker error?
- Do you have any benchmarks on the time it takes to load a cold parser for the first time?
- I think we can remove this https://github.com/hexpm/hexpm/blob/main/lib/hexpm_web/syntax_highlight.ex#L14.

leandrocp:
Hey @ericmj 

> Can you look at the docker error?

Fixed by updating mdex_native.
 
> Do you have any benchmarks on the time it takes to load a cold parser for the first time?

That depends on network speed because it needs to download the WASM files from CDN. If that's a concern those files can be cached in `Dockerfile` and the app just load it, for example:

```
❯ export LUMIS_DATA_DIR="/tmp/hexpm-test"

❯ rm -rf /tmp/hexpm-test

# do this cache in Dockerfile (download .wasm files)
iex> :timer.tc(fn -> Lumis.Languages.cache(["elixir"]) end)
{4455611,
 {:ok,
  ["/tmp/hexpm-test/parsers/tree-sitter-elixir-0.26.3-3d875f7c1a26af9106752a16a1ab31e3fa22f60de37aba2860edd3277335ad57.wasm"]}}

# do this load on application.ex (warm-up)  
iex> :timer.tc(fn -> Lumis.Languages.load(["elixir"]) end)
{599701, :ok}

# regular user requests
iex> :timer.tc(fn -> Lumis.Languages.load(["elixir"]) end)
{85, :ok}

iex> :timer.tc(fn -> Lumis.highlight!(hexpm_mix_exs, formatter: {:html_inline, language: "elixir", theme: "github_light"}) end)
{19729,
 "<pre class=\"lumis\" .....

❯ du -ha /tmp/hexpm-test/parsers/tree-sitter-elixir-0.26.3-3d875f7c1a26af9106752a16a1ab31e3fa22f60de37aba2860edd3277335ad57.wasm
1.6M    /tmp/hexpm-test/parsers/tree-sitter-elixir-0.26.3-3d875f7c1a26af9106752a16a1ab31e3fa22f60de37aba2860edd3277335ad57.wasm
```

That `Lumis.Languages.load/1` does download (if file is missing) and load so the warmed request is just ~100 microseconds (that last load example).

So the most optimized approach is:

1. Download (cache) all possible languages in Dockerfile
2. Load once on app start (safe and async)

Each parser .wasm varies in size, usually from ~500kb to ~4mb

I can do that change if you're okay with it.

> I think we can remove this https://github.com/hexpm/hexpm/blob/main/lib/hexpm_web/syntax_highlight.ex#L14.

Will do 👍🏻

ericmj:
Yes, I think we have to cache in the docker image. We have a 1 second timeout on highlighting to mitigate parser issues from having too much impact on production, specially as crawlers are hitting a lot `/preview` endpoints, and a number of parser loads hit that deadline even on Apple CPUs.

These are the results from the benchmarking I did of loading parsers:

| Metric                         | Uncached load |
| ------------------------------ | ------------: |
| Parsers tested                 | 115           |
| Successful                     | 114           |
| Median                         | 399 ms        |
| p95                            | 1.260 s       |
| p99                            | 2.680 s       |
| Maximum successful             | 2.817 s       |
| Successful loads over 1 second | 13            |

Precompiled parsers load much faster:

| Cache state          | Median | p95     | Maximum |
| -------------------- | -----: | ------: | ------: |
| Download and compile | 399 ms | 1.260 s | 2.817 s |
| Precompiled in image | 32 ms  | 243 ms  | 547 ms  |

I would rather take this cost at build time than at runtime.

PR #1832, opened 2026-08-14
Select latest releases with stored SemVer keys

ericmj:
Store a sortable `bytea` key and stability flag with each release. A trigger fills both columns, and two indexes cover pure SemVer ordering and Hex.pm's stable-first ordering.

Single-release lookups now use an indexed `LIMIT 1` query. Package lists use one indexed lateral lookup per package. The versionless README redirect uses the same query. Complete release lists still come back to Elixir because those callers need every release; sorting the 2,964-release fixture in Elixir took 0.059 ms.

The schema DDL runs in a transaction with a five-second local lock timeout. The backfill commits 10,000-row batches, then applies the constraint and `NOT NULL` changes in a separate transaction. If finalization can't acquire its lock, the DDL rolls back and the next run resumes after the completed batches. The concurrent indexes remain in their own migration.

## Benchmark setup

The fixture sizes come from the production release-count snapshot taken on 2026-08-13. Each profile has 100 packages: 50 at p50, 40 at p90, 9 at p99, and one at the maximum.

| Profile | p50 | p90 | p99 | Maximum |
| --- | ---: | ---: | ---: | ---: |
| All repositories | 3 | 20 | 96 | 2,964 |
| Public Hex | 3 | 17 | 65 | 895 |

The 2x and 4x profiles multiply every release count. The benchmark ran on PostgreSQL 18.4. Baseline and stored-key measurements used six fresh databases. Every value below is the median of three process medians; p99 is the median of the three process p99 values. Fixture creation and `ANALYZE` ran outside the timers.

Full request timings start at `Plug.Test.conn/2` and end after `HexpmWeb.Endpoint.call/2`. They include routing, controller and context work, database execution and decoding, rendering, and response-body construction. They don't include TCP, TLS, Bandit, reverse proxies, or network latency.

## Latest selection

`Packages.attach_latest_releases/1` was timed with 100 already-loaded package structs. The timer includes its Ecto query, database execution, decoding, mapping, and attachment. Loading the package structs is outside the timer.

| Profile | Scale | Baseline median / p99 | Stored key median / p99 | Median change |
| --- | ---: | ---: | ---: | ---: |
| All repositories | 1x | 7.755 / 9.112 ms | 0.388 / 0.650 ms | -95.0% |
| All repositories | 4x | 32.249 / 35.660 ms | 0.423 / 0.721 ms | -98.7% |
| Public Hex | 1x | 3.985 / 4.943 ms | 0.392 / 0.640 ms | -90.2% |
| Public Hex | 4x | 16.072 / 19.026 ms | 0.442 / 0.760 ms | -97.2% |

All-repository BEAM allocation fell from 12.1 MB to 339 KB at 1x and from 48.5 MB to 339 KB at 4x.

`Releases.latest_version/3` includes the package-name subquery, database execution, decoding, and version return.

| Profile | Percentile | Releases | Baseline median / p99 | Stored key median / p99 | Median change |
| --- | --- | ---: | ---: | ---: | ---: |
| All repositories | p50 | 3 | 0.055 / 0.149 ms | 0.058 / 0.213 ms | +4.3% |
| All repositories | p99 | 96 | 0.165 / 0.349 ms | 0.070 / 0.217 ms | -57.9% |
| All repositories | Maximum | 2,964 | 4.234 / 5.154 ms | 0.401 / 0.619 ms | -90.5% |
| Public Hex | p50 | 3 | 0.056 / 0.152 ms | 0.058 / 0.167 ms | +4.4% |
| Public Hex | p99 | 65 | 0.133 / 0.288 ms | 0.063 / 0.145 ms | -52.5% |
| Public Hex | Maximum | 895 | 1.360 / 1.883 ms | 0.156 / 0.396 ms | -88.5% |

The p50 difference is below the benchmark's useful resolution. The indexed query starts paying off as release history grows.

## Full web request

`GET /packages?search=<fixture>` renders 30 packages and calls the latest-per-package query.

| Profile | Scale | Baseline median / p99 | Stored key median / p99 | Median change |
| --- | ---: | ---: | ---: | ---: |
| All repositories | 1x | 10.890 / 14.986 ms | 2.190 / 3.054 ms | -79.9% |
| All repositories | 4x | 31.132 / 39.997 ms | 3.543 / 5.096 ms | -88.6% |
| Public Hex | 1x | 8.041 / 10.285 ms | 5.289 / 7.266 ms | -34.2% |
| Public Hex | 4x | 16.276 / 20.871 ms | 5.288 / 7.514 ms | -67.5% |

The all-repository 1x request stayed at six database queries. Cumulative query time changed from 2.988 ms to 1.298 ms.

## Controls and migration

Endpoints that load complete release histories didn't improve. API search median changes ranged from +24.0% to -5.9% across the six profiles. Representative package endpoint medians usually moved by single-digit percentages, with one +16.1% result. Their query counts didn't change.

The migration benchmark used 258,347 releases.

| Work | Time or size |
| --- | ---: |
| Columns, trigger, batched backfill, constraints, and `NOT NULL` | 4.1 s |
| Concurrent index migration | 0.3 s |
| Release table before migration | 64 MB |
| Release table after backfill and both indexes | 176 MB |

The final migration computes each backfilled key once instead of twice, so 4.1 seconds is an upper-bound measurement.


lib/hexpm/ecto/version.ex
```diff
@@ -1,6 +1,11 @@
 defmodule Hexpm.Version do
   @behaviour Ecto.Type
 
+  @numeric_identifier 1
+  @alphanumeric_identifier 2
+  @max_numeric_component 9_223_372_036_854_775_807
```
maennchen: I needed to look up what that value is, might be worth documenting.


```suggestion
  # Postgres bigint max (2^63 - 1); must match the check in hexpm_semver_sort_key()
  @max_numeric_component 9_223_372_036_854_775_807
```

maennchen: Correction, we should not allow larger than 14 digits for numeric segments.

See https://github.com/elixir-lang/elixir/commit/c64417d


```suggestion
  @max_numeric_component 99_999_999_999_999
```

priv/repo/migrations/20260814120000_add_release_semver_sort_key.exs
```diff
@@ -0,0 +1,128 @@
+defmodule Hexpm.Repo.Migrations.AddReleaseSemverSortKey do
```
maennchen, line 1: Conceptually a GENERATED column might be a better choice. Not sure how you would do that nicely with a migration that does not block.

maennchen: We could also index the expression itself, which would remove the need for triggers and a stored column.

ericmj: I implemented this and benchmarked it against the stored columns on PostgreSQL 18.4. PostgreSQL uses the expression index to locate each release, but it still evaluates `hexpm_semver_sort_key(version)` for the row returned by each lateral query. Function tracking recorded 100 calls for a 100-package query.

Across three processes, the complete `attach_latest_releases/1` call for 100 already-loaded public-profile packages increased from 0.378 ms to 1.169 ms. An in-process Phoenix web search request for 30 packages, including database work and rendering but excluding network and fixture setup, increased from 4.584 ms to 6.015 ms.

The expression version also computes the key twice on release writes because both indexes contain it. The stored-column backfill took 4.1 seconds for a local 258,347-release fixture, and the indexes are built concurrently. I prefer the one-time migration cost over the recurring read and write cost, so I’m keeping the stored columns and trigger.

ericmj: I've switched this to two `GENERATED ALWAYS AS (...) STORED NOT NULL` columns. [PostgreSQL 14 supports stored generated columns](https://www.postgresql.org/docs/14/ddl-generated-columns.html), so this doesn't require a production upgrade.

The `ALTER TABLE` takes an `ACCESS EXCLUSIVE` lock while PostgreSQL computes both values, rewrites the heap, and rebuilds the existing indexes. Five independent PostgreSQL 14 runs against 258,347 releases and an 84.1 MiB heap took 0.976 to 1.010 seconds, with a 0.983-second median. That times the `ALTER TABLE` itself; fixture creation, `VACUUM`, validation queries, and the two new concurrent index builds were excluded.

Production is a regional 4-vCPU, 32 GB N2 Cloud SQL instance with 300 GB PD-SSD. Accounting for its CPU, regional storage, WAL, and normal load, I expect 3 to 6 seconds of lock time and would budget 10 seconds. The five-second `lock_timeout` only limits how long the migration waits to acquire the lock.

This removes the trigger and batched backfill, and the database guarantees that both stored values match `version`. The two indexes remain in separate concurrent migrations.

PR #1836, opened 2026-08-16
Fix docs navigation layout on tablet view

### LINKED ISSUE #1802: [Bug]: Docs navigation menu is not rendered correctly on tablet view
### Product or component

Hex.pm website or API

### Summary

The documentation navigation menu does not render correctly on tablet-sized screens. The sidebar takes up most of the viewport, leaving very little space for the documentation content.

### What happened?

When viewing the Docs page on a tablet, the navigation menu remains expanded and occupies a large portion of the screen. This results in a cramped reading area and a suboptimal browsing experience.

### What did you expect?

The Docs navigation should adapt to tablet layouts by collapsing into a drawer, accordion, or otherwise providing more space for the documentation content.

### Steps to reproduce

1. Open the Hex Docs page.
2. Switch to a tablet-sized viewport (or use a tablet device).
3. Open any documentation page.
4. Observe that the navigation menu occupies most of the screen width, reducing the available content area.

### Environment / context

N/A

### Logs / screenshots

Image

### Additional context

_No response_

BittuBarnwal7479:
## Summary

This PR improving the responsive layout of the documentation navigation on tablet-sized screens.

The docs navigation now adapts better to smaller viewports, giving the documentation content more available space and improving the reading experience.

## Testing

N/A

## Fixes: #1802


ericmj:
I pushed 63eda69 on top of your commit. It keeps the goal from #1802 but swaps the drawer for an in-flow accordion, and I want to say why.

The drawer forced the panel open at desktop widths with ::details-content, which Firefox added in 143 and Safari in 18.4 (https://caniuse.com/mdn-css_selectors_details-content). The `display: block` fallback can't reveal a closed <details> in older engines, so Firefox ESR 140 and Safari <= 18.3 would have had no docs navigation above 1024px. Testing the drawer also showed page scroll wasn't locked, the close control sat outside the dialog under the backdrop, and the hand-written breakpoint disagreed with Tailwind's `lg:` at fractional zoom.

The new commit renders the nav twice from one partial: the existing sticky sidebar from lg up, and a closed <details> accordion below it with a summary card showing the current page. No JS, no inline CSS, and both renders switch on the same `lg:` utilities. The link list moved into DocsView as data so the two can't drift.

Happy to adjust the accordion (open by default, sticky, etc.) if you'd rather.

ericmj:
I pushed 63eda69 on top of your commit, replacing the drawer with a plain accordion.

The main problem with the drawer was that it depended on `::details-content` to keep the panel open at desktop widths. That only exists in Firefox 143+ and Safari 18.4+ (https://caniuse.com/mdn-css_selectors_details-content), and the `display: block` fallback doesn't work because older engines hide closed `<details>` content in the shadow slot, so on Firefox ESR 140 or Safari 18.3 the docs pages would have had no navigation at all on desktop. There were some smaller issues too: the page behind the drawer still scrolled, the close button was outside the dialog and under the backdrop, and the hard-coded breakpoint didn't line up with Tailwind's `lg:` at fractional zoom levels.

The accordion is simpler. Below lg the nav is a closed `<details>` with a summary showing the current page, above lg it's the same sticky sidebar as before. Both are rendered from one partial and the link list now lives in DocsView so they can't get out of sync. No JS or inline CSS needed.

PR #1837, opened 2026-08-16
Refactor theme toggle functionality and remove unused theme menu

BittuBarnwal7479:
## Summary

Currently, changing the theme requires opening the dropdown menu and then selecting an option. This adds an unnecessary interaction.
This PR allowing users to switch themes directly by clicking the theme icon.

## Testing

N/A

## Screen record:

https://github.com/user-attachments/assets/bc8bcc49-a388-4f31-be78-840d19770ca1

### DIFF
--- a/assets/css/tailwind.css
+++ b/assets/css/tailwind.css
@@ -293,17 +293,6 @@ html[data-theme-preference="system"] [data-theme-toggle] [data-theme-icon="syste
   display: inline-flex;
 }
 
-/* Highlight active choice in theme menu */
-[data-theme-menu] [data-theme-choice] {
-  font-weight: normal;
-}
-
-html[data-theme-preference="light"] [data-theme-menu] [data-theme-choice="light"],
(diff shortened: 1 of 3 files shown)


ericmj:
The downside I see with this is that it's less clear what the icons mean, the sun and the moon are somewhat self explanatory, but the icon that follows the system theme isn't really without the extra text from the dropdown.

BittuBarnwal7479:
> The downside I see with this is that it's less clear what the icons mean, the sun and the moon are somewhat self explanatory, but the icon that follows the system theme isn't really without the extra text from the dropdown.

that's a fair point. i’ve added a tooltip like “Switch to system mode” for extra context. Also, by default the 3rd option is the system mode, so I thought this would make switching faster while still explaining the icon.

ericmj:
Thanks for the PR, but I think the dropdown is a clearer UX and this is only saving one click.

Issue #1839, opened 2026-08-18
[Bug]: swoosh 1.27.1 missing from /packages/swoosh registry index while present in the API


Elmseld:
### Product or component

Infrastructure

### Summary

`repo.hex.pm/packages/swoosh` lists versions only up to 1.27.0. Version 1.27.1 is
absent from the index although the release is published, not retired, and its
tarball is served correctly. Any `mix deps.get` against a lockfile pinning swoosh
1.27.1 fails. The served index object is a cached copy predating the release, and
it appears the purge on publish did not take effect on the shield node.

### What happened?

Projects that took the swoosh 1.27.1 bump can no longer resolve dependencies:

    ** (Mix) Unknown package version swoosh 1.27.1 in lockfile

Without a lockfile entry, resolution fails instead:

    Because "your app" depends on "swoosh 1.27.1" which doesn't match any
    versions, version solving failed.

Decoding the index confirms the version is simply absent — the newest entry is
1.27.0. Meanwhile the release exists and is healthy:

- `GET /api/packages/swoosh/releases/1.27.1` returns 200, `"retirement":null`,
  `"inserted_at":"2026-08-15T06:28:48Z"`
- The tarball is served, and its sha256 matches both the API checksum and the
  `outer_checksum` recorded in our lockfile
  (`9e732f469259e5e84364967b6e0b5d5562adf35bd77576b92a6f3753de6ae9e1`)

This is not client-side: it reproduces with plain `curl`, with no Hex involved,
and is identical with and without `Accept-Encoding: gzip`. It also reproduces via
the jsdelivr mirror, which fronts the same origin. We checked all 133 hex packages
in our lockfile against the registry — swoosh 1.27.1 is the only one missing.
`/versions` itself is fresh (`age: 0`).

### What did you expect?

That `/packages/swoosh` would include 1.27.1 shortly after the release was
published on 2026-08-15 06:28 UTC, and that `mix deps.get` would resolve a
lockfile pinning that version.

### Steps to reproduce

1. Confirm the index tops out at 1.27.0:

       curl -s https://repo.hex.pm/packages/swoosh | gunzip \
         | grep -ao '1\.2[0-9]\.[0-9]*' | sort -uV | tail -3

2. Confirm the release nevertheless exists:

       curl -s https://hex.pm/api/packages/swoosh/releases/1.27.1 | head -c 200

3. In any project, pin `{:swoosh, "== 1.27.1"}` and run `mix deps.get`.
[...]

Issue #1838, opened 2026-08-17
Registry object for package "starter" stuck at 0.3.2, not reflecting published 0.4.0/0.4.1


jamilabreu:
## Summary

The registry object served at `https://repo.hex.pm/packages/starter` (the endpoint `mix deps.get` actually resolves dependencies against) has not been rebuilt/purged since `0.3.2` was published on 2026-08-06, even though two newer versions have since been published successfully and are visible everywhere else (hex.pm website, the JSON API, `mix hex.info`).

## Timeline

- `0.4.0` published: `2026-08-17T03:31:18Z`
- `0.4.1` published: `2026-08-17T03:50:45Z`
- `0.4.1` republished with `mix hex.publish --replace --yes` (still within the 1-hour window): `2026-08-17T04:28:10Z` (this is `0.4.1`'s `updated_at`)
- As of `2026-08-17T22:08:55Z` (~18.5 hours after the replace-publish), the registry object is still unchanged.

## Evidence

`https://hex.pm/api/packages/starter/releases/0.4.1` (JSON API — correct):
```
inserted_at: 2026-08-17T03:50:45.845027Z
updated_at:  2026-08-17T04:28:10.621085Z
```

`https://repo.hex.pm/packages/starter` (registry endpoint — stale):
```
$ curl -sI https://repo.hex.pm/packages/starter
last-modified: Thu, 06 Aug 2026 18:35:27 GMT   <- matches when 0.3.2 was published, not 0.4.0/0.4.1
age: 63628
etag: "b422a3e54b7d2337955551d2af7a7c8b"

$ curl -s https://repo.hex.pm/packages/starter | gunzip | strings | grep -E '^0\.'
0.1.0
0.1.1
0.1.2
0.2.0
0.2.1
0.3.0
0.3.1
0.3.2
```
`0.4.0` and `0.4.1` are missing from this list.

## Impact

`mix deps.get` against `{:starter, "~> 0.4.1"}` fails for every consumer with:
```
Because "your app" depends on "starter ~> 0.4.1" which doesn't match any versions, version solving failed.
** (Mix) Hex dependency resolution failed
```
even though the versions are fully published and downloadable (confirmed `https://repo.hex.pm/tarballs/starter-0.4.1.tar` returns 200).

## What I've already tried

- Cleared local Hex caches (`~/.hex/cache.ets`, cached tarballs) — no effect, confirming this is server-side, not a local cache issue.
[...]

PR #1841, opened 2026-08-18
Purge Fastly from an Oban job and verify the purge took

ericmj:
Fastly accepted every purge request during the incident behind #1838 and
#1839 but the entries on the IAD shield survived, and nothing noticed until
users reported stale registry objects days later. Purges now run in a
Hexpm.CDN.PurgeWorker job on the purge queue with the surrogate keys and a
list of verify targets (URL and the ETag the store returned, nil when the
object was deleted). The job purges twice two seconds apart as Fastly
recommends, waits, then fetches every target through the CDN and compares
the served ETag. Fastly routes by anycast so the direct fetch lands on the
nearest POP; for the repository service the check also asks the IAD shield
and one POP per other continent (stockholm, tokyo, sydney, sao paulo,
johannesburg) in parallel through the Compute service's hex-cache-probe
header, which answers from that POP's cache over Fastly's shield network.
Every answer carries x-cache-served-by so a stale result names the caches
involved. A stale target anywhere is purged again; after the configured
rounds the job raises Hexpm.CDN.PurgeVerificationError, so Oban retries it
and Sentry reports it through the Oban integration. Queued jobs for the
same service are absorbed into the running one, up to Fastly's 256 keys per
request. Two targets for one URL keep the later ETag, and a stale target is
dropped rather than re-purged when a newer job for its URL exists, since
that job carries the ETag the store holds now.

Measured on staging with the same object shape: after a purge the shield
served the new object within 77-383 ms in 17 of 18 samples and 1.8 s in one,
and every probed POP within 2 s, so the two-second gap between the purges
and the two-second grace before the check sit above the observed tail.

Private repository objects are checked with a short-lived JWT for that
repository, so the signing key moves to the shared runtime config. hexdocs
purges verify the uploaded index.html and the package sitemap at the
package subdomain, where the CDN serves it.

Hexpm.CDN.Fastly.purge_key sends one request, retries 429 and 5xx, and
returns the status and body on failure instead of raising. New telemetry
covers purge requests, verifications per POP and purge jobs and is exported
through PromEx. The purge queue concurrency comes from
HEXPM_OBAN_PURGE_CONCURRENCY.

Registry builds still purge synchronously under the registry lock, twice
and raising on a rejected purge, as they did.


lib/hexpm/cdn/purge_worker.ex
```diff
@@ -0,0 +1,252 @@
...
+  `Hexpm.CDN.PurgeVerificationError` and Oban retries it later.
+
+  Purges queue up faster than they run during publish bursts, so on start a
+  job absorbs the available jobs for the same service: their keys and
+  targets are merged into this job's arguments (persisted, so a retry keeps
+  them) and the absorbed jobs are cancelled. Fastly accepts up to 256 keys
+  per purge request, so a job never carries more than that, and it looks at
+  no more than 256 candidates, since each brings at least one key. Two
+  targets for the same URL keep the later one's ETag, since that is the
+  object the store holds now.
```
maennchen, line 22: You now have a worker that essentially executes Task.async_stream twice nested. Would it make more sense to just increase worker capacity so that they get executed more quickly? I used oban before to run thousands of jobs a second, so that should be possible and simpler.

ericmj: The nesting is gone: the behaviour is batch-shaped now, `verify(service, targets)`, so the worker makes one call and the Fastly module runs a single task stream over targets times POP checks with one concurrency cap. Capacity alone wouldn't replace the absorption though. The queue already runs at concurrency 5 and Oban would keep up fine, but N separate jobs send 2N purge requests and each sleeps ~4 s for propagation before verifying, while a merged job sends 2 requests carrying up to 256 keys, which is the batch shape Fastly's surrogate-key endpoint is built for. Merging also collapses two writes of the same URL into one target with the later ETag; as separate jobs the older one can never verify and only finds that out after a stale round.

lib/hexpm/cdn/fastly.ex
```diff
@@ -1,33 +1,191 @@
...
+      fn {target, pop, headers} -> {target, pop, check(target, pop, headers)} end,
+      max_concurrency: @verify_concurrency,
+      timeout: @probe_timeout + 5_000,
+      on_timeout: :kill_task
+    )
+    |> Enum.zip(checks)
+    |> Enum.map(fn
+      {{:ok, result}, _check} -> result
+      {{:exit, reason}, {target, pop, _headers}} -> {target, pop, {:error, {:exit, reason}}}
+    end)
```
maennchen, line 88:
```suggestion
    |> Task.async_stream(
      fn {target, pop, headers} -> {target, pop, check(target, pop, headers)} end,
      max_concurrency: @verify_concurrency,
      timeout: @probe_timeout + 5_000,
      on_timeout: :kill_task,
      ordered: false,
      zip_input_on_exit: true
    )
    |> Enum.map(fn
      {:ok, result} -> result
      {:exit, {{target, pop, _headers}, reason}} -> {target, pop, {:error, {:exit, reason}}}
    end)
```

PR #1842, opened 2026-08-18
Build the registry in an Oban job and purge it through the purge job

ericmj:
Publishing built the package object under the registry advisory lock inside
the request and rebuilt names and versions in a fire-and-forget task, so a
slow lock held the publish and a crash in the task was lost.
Hexpm.Repository.RegistryWorker now runs the build on a registry queue:
package builds at priority 0, repository index builds at 1, full builds at
2. Publishing enqueues the build once the tarball is in the store, so the
registry never names a release without one; reverts, retirements,
advisories and admin removals enqueue inside the transaction that changed
the data. A job waits up to a minute for the registry lock and snoozes
otherwise, so builds queued behind a full build neither hold a connection
nor spend attempts on waiting.

Queued work is consolidated when a job starts, inside its transaction: a
package job takes over up to a hundred queued package jobs of its
repository and builds them with one round of queries, concurrent uploads
and one purge; a full build cancels every queued job of its repository;
repository and package_delete jobs cancel queued jobs with the same
arguments. A failed build rolls the cancellations back. A package_delete
leaves the object alone when a package with that name exists again, since
its own build owns the object by then.

RegistryBuilder no longer locks, purges or logs. Every builder returns the
surrogate keys to purge and the objects to verify (store key and ETag), and
the worker hands them to Hexpm.CDN.purge, which enqueues the purge job in
the same transaction. Full builds verify names and versions only.

Telemetry for builds by type and result is exported through PromEx. The
registry queue concurrency comes from HEXPM_OBAN_REGISTRY_CONCURRENCY. The
hex client tests publish and resolve straight after, so the hex environment
runs Oban inline.


lib/hexpm/repository/registry_worker.ex
```diff
@@ -0,0 +1,271 @@
...
+    query =
+      from(j in Oban.Job,
+        where: j.worker == ^worker and j.state == "available" and j.id != ^id,
+        where: ^filter,
+        order_by: j.id,
+        lock: "FOR UPDATE SKIP LOCKED"
+      )
+
+    query = if limit, do: from(j in query, limit: ^limit), else: query
+    Repo.all(query)
```
maennchen, line 248: Why are you loading the query here?

`Oban.cancel_all_jobs` takes a query, so instead of materializing and chunking, we could just hand the query, no?

ericmj: We need the rows anyway. `siblings/3` selects them `FOR UPDATE SKIP LOCKED` so consolidations on different nodes split the queue instead of waiting on each other, and the package branch reads `package_id` off them to build the batch, so cancelling by id costs nothing extra and cancels exactly what we absorbed. Passing the filter to `cancel_all_jobs` would re-run it at update time: the UPDATE waits on rows another node holds, and it can cancel a job enqueued after our select that isn't in the batch, so nobody builds that package. The chunking is only there because full consolidations have no limit.

maennchen (approved):

PR #1849, opened 2026-08-21
Skip symlinks when walking unpacked tarballs and raise storage transfer concurrency

ericmj:
The diff worker timed out all 5 attempts (270s each) generating the diff for svx 0.2.0 → 0.3.0 (HEXPM-D6). svx 0.2.0 ships a vendored example app whose `_build` contains symlink cycles, and both the diff generator and the preview uploader walked unpacked tarballs with `Path.wildcard`, which descends into symlinked directories, and `File.regular?`, which resolves symlinks. The package's 1,171 regular files walked as 54,691 paths on a prod worker, and every extra path meant a git spawn plus a sequential GCS PUT.

Three changes: a shared `lstat`-based walk that never follows symlinks (used by diff generation and preview uploads), diff pieces uploaded through `Task.async_stream` instead of one blocking PUT per changed file, and transfer concurrency raised from 10 to 32 everywhere a job streams objects (diff pieces, preview files, hexdocs pages, registry objects, `delete_many`), with the two diff tarball downloads now fetched in parallel. A concurrency sweep on a prod worker scaled linearly to 50 (7.7–8.2s at 10, 2.4–2.6s at 32, 1.5–1.6s at 50 for 1,169 pieces, within 11% of ideal against the 63ms mean per-object latency), so the limit is pool budget rather than throughput: the GCS Finch pool grows from 50×2 to 50×8 to cover the worst case of 10 heavy jobs plus 2 registry jobs at 32 concurrent requests each.

Phase timings for the failing job, measured on a prod worker pod (4 schedulers) by running the generator logic through the release's rpc on the real job args, single runs; totals also include ~0.3s of tarball downloads:

| configuration | tree walk | unpack | diff loop | piece uploads | job total |
|---|---|---|---|---|---|
| main | 29.0s (54,691 paths) | 20.8s | 254.8s | did not finish | >270s, timed out ×5 |
| symlink-safe walk | 0.02s (1,171 files) | 22.1s | 2.9s | 74.4s (sequential) | 99.8s |
| + concurrent uploads (10) | 0.02s | 22.1s | 2.9s | 7.0s | 32.4s |
| + concurrency 32 (this PR) | 0.04s | 22.3s¹ | 2.3s | 2.7s | ~27s¹ |
| + hexpm/hex_core#220 | 0.04s | 0.5s | 2.3s | 2.7s | **5.7s** |

¹ derived by combining this row's measured phases with the unpack timing from the earlier run; every other total is the sum of one run's phases.

The remaining unpack cost is the same symlink-following wildcard inside hex_core's post-extract mtime pass, fixed in hexpm/hex_core#220 (now merged); the last commit bumps mix.lock to that merge commit, so the 5.7s row is what this PR ships. Preview uploads gain the same protections: a symlink-cycle package can no longer explode the uploaded file list, and a 1,000-file preview upload drops from ~70s to ~2s.

The end-to-end tests build tarballs that actually contain symlink entries by rebuilding `contents.tar.gz` with erl_tar and recomputing the checksums, since current `hex_tarball.create` rejects symlinks while tarballs published by old Hex clients still contain them.


lib/hexpm/diff/generator.ex
```diff
@@ -198,12 +217,8 @@ defmodule Hexpm.Diff.Generator do
...
   defp executable?(mode), do: band(mode, 0o111) != 0
 
-  defp tree_files(directory) do
-    directory
-    |> Path.join("**")
-    |> Path.wildcard(match_dot: true)
-    |> Enum.filter(&File.regular?(&1, raw: true))
-    |> Enum.map(&Path.relative_to(&1, directory))
+  defp regular_file?(path) do
+    match?({:ok, %File.Stat{type: :regular}}, File.lstat(path))
```
maennchen, line 221: Why not `File.regular?`?

maennchen: Sorry, the PR description covers this.

maennchen (approved):

PR #1865, opened 2026-08-28
Render additional package documentation files

### LINKED ISSUE #1569: Support rendering additional package documentation files
We recently added `README` rendering to Hex.pm package detail pages. It would be useful to extend this to other common project documentation files so package users can find important information directly on Hex.pm.

Suggested files to support:

```text
README
LICENSE
SECURITY
CHANGELOG
SUPPORT
ACKNOWLEDGMENTS
THREAT_MODEL
```

We should also consider looking for these files in customary locations, not only the package root. Possible search locations:

```text
/
.github/
.gitlab/
```

The package detail page could expose these documents as separate tabs or sections, similar to how GitHub does it. For example:

```text
README | Changelog | License | Security | Support | Threat Model | Acknowledgments
```

Motivation:

Many packages already include important metadata and project policies in these conventional files, but users currently need to leave Hex.pm and inspect the source repository or package contents manually. Rendering these files directly would make it easier to:

* review license information
* find security reporting instructions
* understand support expectations
* inspect changelog/release notes
* discover threat models or security design notes
* surface project acknowledgments and contributor information

This would also help encourage maintainers to publish security-relevant project documentation in standard locations, especially files like `SECURITY.md`, `SUPPORT.md`, and `THREAT_MODEL.md`.

Open questions:

* Which filenames and extensions should be recognized?
* Which directories should be searched?

Ch4s3:
Closes #1569.

Extends README rendering to six more conventional documentation files: Changelog, License, Security, Support, Acknowledgments, Threat Model.

## Recognition

Root of the tarball only, no `.github/`. Bare filename, `.md`, `.markdown`, or `.txt`, case-insensitive, with `.md` preferred and a bare file beating `.txt` (matches README's existing preference order). `Acknowledgments` also matches the British spelling, `Threat Model` also matches `THREATMODEL` and `THREAT-MODEL`. All of this lives in `HexpmWeb.Docs.Files`, one table to extend for future file types.

## Rendering

`ReadmeController` becomes a general doc renderer on the `readme.` host: `/:name/:version/:kind` alongside the existing `/:name/:version`. Same pipeline as README today (Earmark, sanitizer, URL rewriter, Makeup), same CSP, same iframe sandbox. Bare and `.txt` files render as escaped `<pre>`, which covers the majority of LICENSE files.

Fetched documents over 2MB don't render inline; the page links out to HexPreview instead. The underlying HTTP client (`Hexpm.HTTP`) now supports a `max_body_size` option that aborts the fetch mid-stream rather than buffering the whole response first.

## Package page

First tab renamed from "Readme" to "Documentation". A file sidebar (desktop) and dropdown group (mobile) list the available files, populated client-side from a `doc-files` postMessage the iframe already had the data to send. A package with only a README looks exactly like it does today, no extra chrome. Deep links to a file the package doesn't publish show an explanation instead of a blank page.

## Testing

New tests for the filename table, the doc controller, and the package page routes/tabs. Existing README behavior is unchanged. `mix test` is green except for 9 pre-existing OAuth failures unrelated to this change.

### DIFF
--- a/assets/js/app.js
+++ b/assets/js/app.js
@@ -123,7 +123,7 @@ window.liveSocket = liveSocket;
 var readmeFrame = document.getElementById("readme-frame");
 
 window.addEventListener("message", function (event) {
-  if (!event.data || !readmeFrame) return;
+  if (!event.data || !readmeFrame || event.source !== readmeFrame.contentWindow) return;
 
   if (
     event.data.type === "readme-height" &&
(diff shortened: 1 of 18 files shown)


Ch4s3:
Closing this one. It was built against a fork of main that had drifted ~300 commits behind hexpm/hexpm — in that gap, README rendering moved from an external CDN fetch to the in-process `Hexpm.Preview` context (mdex instead of Earmark, no external CDN dependency), and a generic "Files" tab (`HexpmWeb.PreviewLive`) was added to the package page that already lets people browse to CHANGELOG.md/LICENSE/etc. in a release.

Given that, I don't think reviving this exact implementation is the right move — it would either fight or duplicate what's already there. Re-scoping as a smaller addition on top of the current Files tab / Preview context (e.g. recognizing and surfacing conventional doc files rather than rebuilding rendering) and will open a fresh PR against current main if that still makes sense after re-reading the issue.

PR #1871, opened 2026-08-30
Add copy button to documentation code blocks

### LINKED ISSUE #1846: [Feature]: Copy Code button copies Markdown formatting
### Product or component

Hex.pm website or API

### Problem statement

Code examples in the documentation do not have a Copy Code button. Users have to manually select and copy the code, which can be inconvenient, especially for larger code examples.

### Proposed solution

Add a Copy Code button to code blocks in the documentation.

The button should copy only the code content to the user's clipboard and provide some visual feedback, such as changing the button text to "Copied" after clicking.

### Why is this useful?

A Copy Code button would make it easier for users to quickly copy examples and use them in their own projects. It would also provide a more convenient experience when working with longer code examples.

### Alternatives considered

Users can manually select the code and copy it, but this is less convenient and can accidentally include unwanted formatting or text.

### Additional context

Before:

Image

After:

Image

peguimasid:
Fixes #1846.

## Summary

Adds a copy button to fenced/indented code blocks on `/docs/*`. It copies the source (`literal`), not the markdown fences or highlighted HTML, and reuses the existing `CopyButton` hook (icon + “Copied!” tooltip).

Blog and policy pages share this markdown engine but do not get the button. Easy to turn on later if we want it there too.

`nohighlight` / shell-session blocks keep the button on purpose — those are the commands people are most likely to copy.

## Testing

`mix test test/hexpm_web/markdown_engine_test.exs test/hexpm_web/controllers/docs_controller_test.exs`

Checked `/docs/usage` in the browser: button appears, clipboard is the code only (no fences).

## Screenshots

image

## Open question

Copy includes `$ ` prompts as written. Other docs often strip `$ ` (and output) so paste works in a terminal. I’d leave it as-is here: a lot of these blocks are session transcripts, not single pasteable commands. Happy to follow up if we want a narrower strip (e.g. only on one-line `$ …` blocks).


ericmj:
Thank you! 💜

PR #1875, opened 2026-08-30
Render additional package documentation files

### LINKED ISSUE #1569: Support rendering additional package documentation files

Ch4s3:
Closes #1569.

Extends README rendering to six more conventional documentation files: Changelog, License, Security, Support, Acknowledgments, Threat Model (per [1569](https://github.com/hexpm/hexpm/issues/1569). Reuses the existing `Hexpm.Preview` context and `HexpmWeb.Readme.Renderer` — the same infrastructure that already renders README — rather than adding anything new to fetch or render files.

## Finding the files

I check the root of the tarball only. Bare filename, `.md`, `.markdown`, or `.txt`, case-insensitive, with `.md` preferred and `.txt` beating a bare file — matching README's existing filename priority exactly. Acknowledgments also matches the British spelling, Threat Model also matches `THREATMODEL` and `THREAT-MODEL`. They all sit in `Hexpm.Docs.Files`, one table to extend for future file types.

## Backend

`Hexpm.Preview.readme/3` is a thin wrapper around a new `doc/4`, which reads any of the above file types the same way (`Bucket.get_file_list` → resolve a filename → `Bucket.get_file`). `ReadmeController` reads an optional `?kind=` query param rather than a new route — same precedent as the existing `?token=` on the private path, and avoids a routing collision a path-based `:kind` segment would hit against the existing all-dynamic `/:repository/:name/:version` route. `HexpmWeb.ReadmeToken` is untouched; it already authorizes a whole release, which covers every kind.

The parent package page gets one route per file type (kind) (`/packages/:name/:version/changelog`, etc.) generated from `Hexpm.Docs.Files.kinds/0`. These have to be literal, not a dynamic `:kind` param, for the same routing-collision reason as above — verified with `Phoenix.Router.route_info/4` that the literal routes don't shadow or get shadowed by any existing route.

## Package page

First tab renamed from "Readme" to "Documentation". The package controller now computes which kinds a release actually has (via the same `Bucket.get_file_list` call the existing Files tab already makes) and renders the sidebar (desktop) and mobile dropdown entries server-side — no client-side JS anywhere in this change. A package with only a README renders identically to today. A deep link to a kind the package doesn't publish shows an explanation in place of the iframe, without an unnecessary fetch.

#### Light Desktop
image

#### Dark Desktop
image

####Light Mobile
image

#### Dark Mobile
image

### DIFF
--- a/lib/hexpm/docs/files.ex
+++ b/lib/hexpm/docs/files.ex
@@ -1,0 +1,119 @@
+defmodule Hexpm.Docs.Files do
+  @moduledoc """
+  Recognition table for conventional documentation files in a release.
+
+  Matched case-insensitively at the tarball root, preferring .md, .markdown,
+  .txt then bare -- Preview's existing README priority.
+  """
+
+  @type resolved :: %{atom => String.t()}
+
+  @kinds_and_labels [
+    readme: "Readme",
+    changelog: "Changelog",
+    license: "License",
+    security: "Security",
+    support: "Support"
+    # ACKNOWLEDGMENTS and THREAT_MODEL were in none of the 2,680 packages
+    # surveyed on #1569; add a kind here and its basenames below for one.
+  ]
+
+  @kinds Keyword.keys(@kinds_and_labels)
+
+  @basenames %{
+    readme: ["README"],
+    changelog: ["CHANGELOG"],
+    license: ["LICENSE"],
+    security: ["SECURITY"],
+    support: ["SUPPORT"]
+  }
+
+  @extensions ["md", "markdown", "txt", ""]
+
+  @extension_rank @extensions |> Enum.with_index() |> Map.new()
+
+  @basename_to_kind for {kind, names} <- @basenames, name <- names, into: %{}, do: {name, kind}
+
+  @doc "Recognized kinds, in canonical order."
+  @spec kinds() :: [atom]
+  def kinds(), do: Keyword.keys(@kinds_and_labels)
+
+  @doc "Label for a kind, falling back to its string form."
+  @spec label(atom) :: String.t()
+  def label(kind), do: Keyword.get(@kinds_and_labels, kind, to_string(kind))
+
+  @spec parse_segment(term) :: atom | nil
+  for kind <- @kinds do
+    def parse_segment(unquote(Atom.to_string(kind))), do: unquote(kind)
+  end
+
+  def parse_segment(_segment), do: nil
+
+  @doc """
+  Every kind resolvable from `files`, as kind => filename. Malformed input
+  resolves to nothing rather than raising; `files` comes from the tarball index.
+  """
+  @spec resolve_all(term) :: resolved
+  def resolve_all(files) when is_list(files) do
+    files
+    |> Enum.filter(&is_binary/1)
+    |> Enum.reject(&String.contains?(&1, "/"))
+    |> Enum.flat_map(fn filename ->
+      case classify(filename) do
+        {kind, ext} -> [{kind, ext, filename}]
+        :nomatch -> []
+      end
+    end)
+    |> Enum.group_by(&elem(&1, 0))
+    |> Map.new(fn {kind, matches} ->
+      {_kind, _ext, filename} =
+        Enum.min_by(matches, fn {_, ext, name} -> {@extension_rank[ext], name} end)
+
+      {kind, filename}
+    end)
+  end
+
+  def resolve_all(_files), do: %{}
+
+  @doc "Kinds present in a `resolve_all/1` result, in canonical order."
+  @spec present_kinds(resolved) :: [atom]
+  def present_kinds(resolved) do
+    Enum.filter(@kinds, &Map.has_key?(resolved, &1))
+  end
+
+  @doc """
+  Kinds to list in navigation: every kind in `resolved`, plus `active` so a
+  deep link to a file the release lacks still shows as the current page.
+  """
+  @spec nav_kinds(resolved, atom) :: [atom]
+  def nav_kinds(resolved, active) do
+    resolved |> Map.put_new(active, nil) |> present_kinds()
+  end
+
+  @spec resolve(atom, term) :: String.t() | nil
+  def resolve(kind, files) when kind in @kinds do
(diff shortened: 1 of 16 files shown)


ericmj (commented):
This looks great, only a couple of comments.

lib/hexpm_web/package_layout_assigns.ex
```diff
@@ -90,12 +101,25 @@ defmodule HexpmWeb.PackageLayoutAssigns do
           Keyword.get_lazy(opts, :dependants_count, fn ->
             Packages.count_dependants(repositories, package)
           end)
+        end,
+      doc_kinds:
```
ericmj: This makes page rendering wait for preview storage, including on the Versions page where we only need the filenames for the mobile dropdown. I'd like the preview Oban job to store the recognized files in a separate table keyed by release ID, which the page can read from PostgreSQL. I'm fine with showing the previous files while reprocessing and replacing the row after the upload succeeds.

lib/hexpm_web/templates/package/show.html.heex
```diff
@@ -1,58 +1,54 @@
-<% description = @package.meta.description %>
+<% sidebar_kinds = Hexpm.Docs.Files.nav_kinds(@doc_kinds, @doc_kind) %>
 
-<.package_layout {@package_layout} active_tab={:readme} version_pinned?={@version_pinned?}>
+<.package_layout
+  {@package_layout}
+  active_tab={:readme}
+  version_pinned?={@version_pinned?}
+  doc_kind={@doc_kind}
```
ericmj, line 7: The version picker drops the selected document: switching versions from a Changelog page takes you to the README. Please pass `doc_kind` through to the version links so they preserve `/changelog`, `/license`, etc.

lib/hexpm_web/controllers/package_controller.ex
```diff
@@ -262,6 +279,13 @@ defmodule HexpmWeb.PackageController do
     )
   end
 
+  defp doc_title(package, :readme), do: package.name
+  defp doc_title(package, doc_kind), do: "#{package.name} · #{Files.label(doc_kind)}"
+
+  # Absolute because this also feeds `og:url`.
+  defp doc_canonical_url(package, release, doc_kind),
```
ericmj, line 288: This also changes the canonical URL for existing README pages to a versioned URL, including when visiting the unversioned package page. Please keep the package-level canonical for README pages and use the document URL for the additional kinds.

lib/hexpm_web/package_layout_assigns.ex
Ch4s3: How do you want to backfill old packages?

Ch4s3: We could have a job that you run once, or do them on demand as pages load.

lib/hexpm/docs/files.ex
```diff
@@ -0,0 +1,119 @@
...
+  @type resolved :: %{atom => String.t()}
+
+  @kinds_and_labels [
+    readme: "Readme",
+    changelog: "Changelog",
+    license: "License",
+    security: "Security",
+    support: "Support"
+    # ACKNOWLEDGMENTS and THREAT_MODEL were in none of the 2,680 packages
+    # surveyed on #1569; add a kind here and its basenames below for one.
```
maennchen: I would like to include them nonetheless. By specifying what files packages can represent, there's hopefully also maintainers adding it.

Specifically it would be good if Nifs (like `explorer`) did implement `ACKNOWLEDGMENTS` to represent non-hex dependencies and generally packages to declare their intentions around a threat model since that is also what automated scanners will be looking at.

Ch4s3: works for me

lib/hexpm_web/package_layout_assigns.ex
ericmj: Good question, lets do it with a job that backfills them once.

Ch4s3:
```elixir
Hexpm.AdminTasks.backfill_doc_files()
```

ericmj:
Deployed and backfill has run. This looks great, thank you @Ch4s3!

Ch4s3:
@ericmj always happy to find something I can contribute!

maennchen:
I think we should rethink the column widths. On a huge screen, best case, we get around 85 characters in a line. 120 would be nice. worst case we get around 40

https://hex.pm/packages/oidcc/3.9.0/license

Screenshot 2026-09-22 at 22 40 11

https://hex.pm/packages/oban

Screenshot 2026-09-22 at 22 43 33

---

also, https://hex.pm/packages/ash/3.33.9/changelog Is not loading

ericmj:
@maennchen https://github.com/hexpm/hexpm/pull/1937 fixes the ash changelog (the page dropped iframe heights over 100000px), wraps plain text files like the oidcc LICENSE, and moves the files list into a tab row above the content.

PR #1893, opened 2026-09-02
Add blog post on the private package authorization vulnerabilities

ericmj:
Blog post disclosing CVE-2026-75542 and CVE-2026-75554 (GHSA-rfx8-w654-8cpr and GHSA-24ww-j3f4-p49c), fixed on 2026-08-24 in bf0fb9d, 50cffd2 and a8bb81d: what each vulnerability was, that neither was being exploited when found and that earlier use can't be established from the deleted token records, the authorization fix, and the four logging changes (audit archive, entitlement query, identity in the CDN access log, origin and application logs). Organization admins get an email pointing at the post. The second commit adds `Organizations.all_admin_notifiable_emails/1`, the admin counterpart of `Users.all_notifiable_emails/0`, with a `billing_active:` option for the recipient list.

Before merging, set the post date from 2026-09-01 to the publish date; it drives the feed order.

Rendered at http://localhost:4000/blog/private-package-authorization-vulnerabilities:

Rendered blog post

## Announcement email

Once the post is live, send the email from a production iex session. The recipients are the verified primary addresses of every admin of an organization with active billing, each address once:

```elixir
subject = "Hex.pm - Security notice: private package authorization vulnerabilities"

body = """
You're receiving this because you administer an organization on hex.pm.

On 24 August 2026 we found and fixed two vulnerabilities in how hex.pm issues OAuth tokens. The more serious one, CVE-2026-75542, let any hex.pm account exchange an ordinary API key for a token scoped to any organization, including yours, and download that organization's private packages from our CDN. It existed from 18 October 2025 until the fix. The second, CVE-2026-75554, let a user removed from an organization keep downloading its private packages and docs for up to 30 days after removal instead of 30 minutes.

When we found the vulnerabilities, no active token carried a scope for an organization its holder could not access, so neither was being exploited at the time. We can't rule out earlier use: the token records that would show it were deleted by routine cleanup, and our CDN access logs from that period record the client IP, time and path of every download but not the account behind it. For 18 October 2025 to 24 August 2026 we can neither confirm nor exclude that someone outside your organization downloaded its private packages.

The fix needs no change on your side. If you want to assess that period for your organization, we can extract from the CDN logs the client IPs, times and user agents of every download of your organization's private packages, for you to compare against your own team. Reply to this email to request it.

The full account, including what we changed so this question can be answered next time (authorization records archived before deletion, a standing check of every token against membership, and the authorizing token or key recorded on every CDN access-log line), is at https://hex.pm/blog/private-package-authorization-vulnerabilities

Questions to security@hex.pm.

The hex.pm team
"""

Hexpm.Accounts.Organizations.all_admin_notifiable_emails(billing_active: true)
|> Hexpm.AdminTasks.send_email(subject, body)
```

`send_email/3` returns `{:ok, count}` with the number queued, not delivered; delivery goes through the email outbox with Oban retries, and a rerun with the same subject skips the addresses already queued.

### DIFF
--- a/lib/hexpm_web/templates/blog/026-private-package-authorization-vulnerabilities.html.md
+++ b/lib/hexpm_web/templates/blog/026-private-package-authorization-vulnerabilities.html.md
@@ -1,0 +1,118 @@
+## Private package authorization vulnerabilities
+
+<div class="subtitle"><time datetime="2026-09-01T00:00:00Z">1 September, 2026</time> · by Eric Meadows-Jönsson</div>
+
+On 24 August 2026 we found and fixed two vulnerabilities in how hex.pm issues
+OAuth tokens. Both could give an account read access to private packages of an
+organization it was not entitled to. Neither was being exploited when we found
+them: no active token carried a scope for an organization its holder could not
+access. Whether either had been used earlier we cannot say, because the
+records that would settle that had been deleted by routine cleanup. This post
+describes both vulnerabilities, what we could and could not establish, and
+what we changed in authorization and in logging. Organization admins are also
+being notified by email.
+
+### CVE-2026-75542: token exchange for organizations the key holder could not access
+
+Any hex.pm account can create an API key with the `repositories` permission,
+which means "every repository this account has access to". Since 18 October
+2025 such a key could be exchanged, through the OAuth `client_credentials`
+grant, for an access token. The exchange admitted a requested scope if the
+key held `repositories` and the scope string began with `repository:`. The
+organization name after the colon was never resolved against the account's
+memberships, so the key could be exchanged for a token scoped to any
+organization on hex.pm.
+
+Our CDN authorizes private package downloads from the signed token alone,
+without a database lookup. A token created this way was therefore working read
+access to the named organization's private packages for its 30-minute
+lifetime, and it could be created again at will. No request reached the
+application, so nothing there could have refused or recorded it. From
+15 March 2026 the same was true of organization keys.
+
+The advisory is
+[GHSA-rfx8-w654-8cpr](https://github.com/hexpm/hexpm/security/advisories/GHSA-rfx8-w654-8cpr).
+
+### CVE-2026-75554: organization scopes surviving removal from the organization
+
+When a token is created with the plain `repositories` scope, which is what the
+Hex client requests, hex.pm expands it into one scope per organization the
+account is currently a member of, and expands it again on every refresh. A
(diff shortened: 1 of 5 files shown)


lib/hexpm_web/templates/blog/026-private-package-authorization-vulnerabilities.html.md
```diff
@@ -0,0 +1,118 @@
...
+   the secret material removed. The archive records who was granted which
+   scopes, when, and through which key or session.
+2. **A standing entitlement check.** A query over the archive and the live
+   records lists every token whose organization scopes were not covered by
+   membership at the time it was created, judged against the membership
+   history in the organization audit log. It is the check reported above,
+   and it can be run again under a different rule later.
+3. **Identity in the CDN access log.** Each repo.hex.pm access-log line now
+   ends with the token or API key that authorized the request, so a private
+   package download can be attributed to the account behind it.
```
maennchen: I assume it ends with the token / key ID (`kid` claim or something) and not the actual token, right?

lib/hexpm_web/templates/blog/026-private-package-authorization-vulnerabilities.html.md
```diff
@@ -0,0 +1,118 @@
...
+
+The fix needs no change on your side. If you want to assess the period
+before the fix for your organization, we can extract from the CDN logs the
+client IPs, times and user agents of every download of your organization's
+private packages between 18 October 2025 and 24 August 2026, for you to
+compare against your own team. Email [security@hex.pm](mailto:security@hex.pm)
+from an admin account of the organization.
+
+We apologize to the organizations that trust hex.pm with their private
+packages.
```
maennchen: This does not sit quite right with me. Bugs and vulns are a normal part of software engineering. You remediated quickly, publish a good and transparent report etc. Apologizing signals guilt, and I don't think there is any.

Maybe we could write something like this instead?

> We take the trust organizations place in hex.pm seriously. When vulnerabilities are found, we will fix them quickly, investigate their impact, and be transparent about what we know and what we cannot establish.

lib/hexpm_web/templates/blog/026-private-package-authorization-vulnerabilities.html.md
ericmj: Correct, will tweak the wording.

maennchen (approved):

PR #1923, opened 2026-09-17
Fix unreadable reset authenticator dialog text in dark mode

### LINKED ISSUE #1909: [Bug]: [Two-factor authentication reset dialog has unreadable text in dark mode
### Product or component

Hex.pm website or API

### Summary

The reset authenticator app confirmation dialog appears to be missing dark-mode text styles, making its title and message difficult to read.

### What happened?

When I open the “Reset Authenticator App?” dialog with the dark theme enabled, the title and explanatory text are rendered in very dark colors against the dark modal background.

The buttons and close icon remain visible, but the main content has very low contrast.

### What did you expect?

The dialog title and message should use text colors with sufficient contrast against the dark background.

### Steps to reproduce

1. Sign in to hex.pm.
2. Enable the dark theme.
3. Open the account security page at `/dashboard/security`.
4. Under “Authenticator App,” select “Setup New App.”
5. Observe the title and message in the confirmation dialog.

### Environment / context

Component: Hex.pm website
Theme: Dark
OS: macOS
Browser: 152.0.7977.77 (Official Build) (arm64)

### Logs / screenshots

Image

### Additional context

I may be missing some context, but the reusable confirmation modal appears to set light-theme text colors
without corresponding dark-theme variants for its custom heading and message.

dajiaohuang:
## Summary

The reset authenticator dialog was unreadable in dark mode — its heading and message rendered as dark text on the dark modal background.

The shared confirmation modal renders its title through the `:header` slot, which replaces the modal's own title element. That default title already carried a dark-mode color, so content passed through the slot has to supply its own; the component's heading and message had no dark variants.

Fixes #1909.

## Changes

- `lib/hexpm_web/components/confirmation_modal.ex` — add `dark:text-white` to the heading and `dark:text-grey-300` to the message.
- `test/hexpm_web/controllers/dashboard/security_controller_test.exs` — regression test asserting the rendered reset authenticator modal carries both dark-mode classes on its heading and message.

Both dialogs built on this component (reset authenticator, disable two-factor) are fixed by the change.

## Test plan

- `mix format --check-formatted`
- `mix compile --warnings-as-errors`
- `mix test` — 3439 passing
- Confirmed the new test fails against the unpatched component

## Out of scope

Four dialogs build their own modal header instead of using this component and have the same missing dark variants: `add_email_modal.ex`, `delete_email_modal.ex`, `revoke_key_modal.ex`, `revoke_session_modal.ex`. Happy to send a follow-up if useful.


maennchen:
Based on the fact that I got 5 notifications of PRs in as many projects in less than 30 minutes, this looks like AI spam. If I want an AI to fix stuff, I'll ask my own. Closing.

PR #1924, opened 2026-09-18
Update Elixir logo

georgeguimaraes:
Elixir got a new logo with the July website redesign. This replaces the old raster wrapped in an SVG (213KB) with the new mark as real paths (3KB), from the official SVG at https://elixir-lang.org/trademarks/


ericmj:
Thank you! 💜

